Issue #1, opened 2019-02-06
Example for dumping to string


repick3:
Do you have a working example of how to dump a schema to string?

PR #2, opened 2019-02-06
Added example showing how to dump out of schema

repick3:
closing #1


lovasoa:
Don't you think this example belongs more to marshmallow than to marshmallow-dataclass?

repick3:
Probably, using the class.Schema() syntax to dump didn't immediately click with me so I thought adding more examples might make it easier for others in the future.

Issue #6, opened 2019-03-14
Using mypy with marshmallow_dataclass


karlb:
When checking a simple test program with mypy
```python
from marshmallow_dataclass import dataclass

@dataclass
class Foo:
    bar: int

Foo(bar=1)
```
I get the following errors
```
$ mypy type_ex.py
type_ex.py:1: error: Cannot find module named 'marshmallow_dataclass'
type_ex.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
type_ex.py:9: error: Unexpected keyword argument "bar" for "Foo"
```

I can easily silence the first error with ` --ignore-missing-imports`, but probably that is related to why I get the second error. Do I have to do any additional mypy setup do get this to validate?

lovasoa:
For the first error: it looks like there is work to do to make a mypy compliant package. PR are welcome ! https://mypy.readthedocs.io/en/latest/installed_packages.html#installed-packages

As for the second error, I think mypy is not clever enough to treat marshmallow_dataclass.dataclass the same as dataclasses.dataclass. You have to use the [`add_schema` decorator](https://lovasoa.github.io/marshmallow_dataclass/html/marshmallow_dataclass.html#marshmallow_dataclass.add_schema) or the [`class_schema` function](https://lovasoa.github.io/marshmallow_dataclass/html/marshmallow_dataclass.html#marshmallow_dataclass.class_schema) for mypy to understand what you are doing.

karlb:
Using `add_schema` works fine. Thanks! I thought it was supposed to work with `marshmallow_dataclass.dataclass` and mypy checking, so I was afraid that I'm missing something important.

Since I'm using `--ignore-missing-imports` anyway and don't have any experience with providing PEP 561 compliant packages, I probably won't provide such a PR anytime soon.

lovasoa:
I am closing this since #10 fixed the first part of this issue, and the second part (*Unexpected keyword argument*) cannot be fixed on our side.

spapinistarkware:
I think this is a fix for the second part:

from mypy.plugins import dataclasses
from mypy.plugin import Plugin


def plugin(version: str):
    return FixMarshamallowDataClassPlugin


class FixMarshamallowDataClassPlugin(Plugin):
    def get_class_decorator_hook(self, fullname: str):
        from mypy.plugins import attrs
        from mypy.plugins import dataclasses

        if fullname == 'marshmallow_dataclass.dataclass':
[...]

PR #10, opened 2019-04-04
Declare package to be typed via to PEP 561

karlb:
Following https://mypy.readthedocs.io/en/latest/installed_packages.html allows users of marshmallow_dataclass to type check their code against the type declarations in marshmallow_dataclass.

While this works towards https://github.com/lovasoa/marshmallow_dataclass/issues/6, it does not solve it, since mypy does not understand how marshmallow_dataclass' `dataclass` decorator works. I don't know how that case can be solved.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,5 +23,6 @@ setup(
     url='https://github.com/lovasoa/marshmallow_dataclass',
     keywords=['marshmallow', 'dataclass', 'serialization'],
     classifiers=CLASSIFIERS,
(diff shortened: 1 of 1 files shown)


lovasoa:
Thank you for this, @karlb !
I'm not an expert, but I'm not sure mypy can be made to understand how the decorator works at all, unfortunately.

I think this can be worked around by using the original dataclass decorator, plus `@add_schema` from this library.

karlb:
> I think this can be worked around by using the original dataclass decorator, plus `@add_schema` from this library.

Yes, I can confirm that this works fine.

Issue #7, opened 2019-03-19
Conda package? marshmallow 3.0?


aldanor:
Hi, and thanks for a nice package :)

A few quick questions - would it be worth pushing this to conda-forge to make it more discoverable? Would be nice to have it available through conda and not just pip.

Also, given that marshmallow 3.0 is bound to be out soon (the stable RCs are already out), would this package be compatible with it?

Thanks!

lovasoa:
There is a marshmallow 3 version available on pypi. It's a pre-release: https://pypi.org/project/marshmallow-dataclass/6.0.0b7/

lovasoa:
you can install the lastest prerelease with:
```
pip install --pre marshmallow-dataclass
```

lovasoa:
As for the conda package, I will happily accept a pull request adding a conda package generation script.

jadkik:
You may want to update the requirements in the `setup.py` file for the marshmallow 3 pre-releases.

I only managed to install it by using `pip install marshmallow-dataclass==6.0.0b9 marshmallow==3.0.0rc5` with a warning `marshmallow-dataclass 6.0.0b9 has requirement marshmallow<3.0,>=2.0, but you'll have marshmallow 3.0.0rc5 which is incompatible.`.

alexlatchford:
Yeah to jump onto @jadkik's comment, thanks for making a pre-release version but it looks like there is a dependency incompatibility that'd be awesome to get ironed out.

lovasoa:
https://pypi.org/project/marshmallow-dataclass/6.0.0rc1/

mivade:
Since Marshmallow 3.0 is out now, can we get an official release?

lovasoa:
Yes :)

mivade:
Thanks!

PR #11, opened 2019-04-23
Update version requirements for marshmallow3 branch

alexlatchford:
Solves #7


lovasoa:
https://github.com/lovasoa/marshmallow_dataclass/blob/783dffd606df06d1184b2a9c64a9224e8fd6aa0d/setup.py#L26

I think we even need a >=3.0

alexlatchford:
Awesome thanks for clearing this up!

PR #17, opened 2019-05-14
Fix error message for unsupported types

karlb:
Before, the right exception was not thrown because formatting the error message caused an exception itself.

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -191,6 +191,11 @@ def class_schema(clazz: type) -> Type[marshmallow.Schema]:
     >>> _, err = class_schema(NeverValid)().load({})
     >>> err
     {'_schema': ['never valid']}
+
+    >>> class_schema(None)  # unsupported type
+    Traceback (most recent call last):
+      ...
+    TypeError: None is not a dataclass and cannot be turned into one.
     """
 
     try:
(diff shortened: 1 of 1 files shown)


lovasoa:
Great, thank you for the PR ! However, it looks like your new test fails on python 3.6
https://travis-ci.org/lovasoa/marshmallow_dataclass/jobs/532382727

<details>
```
File "marshmallow_dataclass/__init__.py", line 195, in __init__.class_schema
Failed example:
    try:
       class_schema(Tuple)  # unsupported type
    except TypeError:
        pass
Expected nothing
Got:
    <class 'marshmallow.schema.Tuple'>
```
</details>

karlb:
Strange, in python3.6 `Tuple` does have a `__name__` attribute while it doesn't in python3.7. Apparently, it can also be converted into a dataclass on python3.6. I can't think of any good example that fails properly on all versions. If you don't have a suggestion, I could either:
* Remove the test
* Use something weird as an int for input, which does not have a `__name__`. But that would be a very unrealistic way to call this function.

lovasoa:
Yes, I would call the function with `None`, and expect an error to be thrown. You can expect errors with :

```
   >>> class_schema(None)
     ...
   TypeError: None is not a dataclass and cannot be turned into one.
```

karlb:
Using None is an excellent idea! Updated the test and it passes CI, now.

lovasoa:
It's published. Thank you very much for your contribution, @karlb !

PR #18, opened 2019-05-15
add support for enums

evanfwelch:
`marshmallow_enum` is part of the marshmallow ecosystem. Accordingly, as a primary user of dataclasses, I would love `marshmallow_dataclass` to support enums. This PR adds that functionality as well as the requisite doctests.

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -34,8 +34,9 @@ Full example::
       })
       Schema: ClassVar[Type[Schema]] = Schema # For the type checker
 """
-
+from enum import Enum, EnumMeta
 import dataclasses
+import inspect
 
 import marshmallow
 import datetime
@@ -44,7 +45,7 @@ import decimal
 from typing import Dict, Type, List, Callable, cast, Tuple, ClassVar, Optional, Any, Mapping, NewType
 import collections.abc
 import typing_inspect
-import inspect
+
 
 __all__ = [
     'dataclass',
@@ -245,7 +246,7 @@ def field_for_schema(
     >>> int_field = field_for_schema(int, default=9, metadata=dict(required=True))
     >>> int_field.__class__
     <class 'marshmallow.fields.Integer'>
-    
+
     >>> int_field.default
     9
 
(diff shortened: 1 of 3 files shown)


Pipfile
```diff
@@ -10,6 +10,8 @@ sphinx = "*"
 marshmallow = ">=2.0,<3.0"
 typing-inspect = "*"
 dataclasses = "*"
+marshmallow-dataclass = "*"
```
evanfwelch: hmm this could be a mistaken addition to `pipfile`, will remove

lovasoa:
In fact, enums are already supported, but only in the [marshmallow3](https://github.com/lovasoa/marshmallow_dataclass/tree/marshmallow3) branch (published as a pre-release to pypi). I did not backport the feature, because I thought marshmallow 3 was going to be released soon, and we could merge it into master. Unfortunately, marshmallow 3 has been a pre-release for several months, and it looks like it won't be released anytime soon.

So enum support is very welcome, but it should be a backport of how it is done in the marshmallow3 branch. Or even better, we may have a single branch (and a single release) that works with both marshmallow 2 and 3. Working with the two in parallel induces a lot of duplicated efforts and confusion.

evanfwelch:
Hey @lovasoa that makes sense, I'm sorry I missed the support in the marshmallow 3 branch. I too have been waiting for marshmallow 3.x for quite some time and when it gets released a lot of my problems will go away. I will take a peek at your marshmallow 3 branch and see if I have any ideas on how to backport or write it once for both versions. Thanks

evanfwelch:
Have a look now, I think this is what you had in mind

lovasoa:
Yep, it's perfect ! You can merge master (and especially 23ce55baee3e2b3cf1fa32dd33c45d519868083e) into your branch, and we'll be ready to merge. Don't forget to add `marshmallow_enum` to the `extra_requires` in `setup.py`.

evanfwelch:
Thanks for the code review, I will update this branch with those changes (I like the `extra_requires` approach too).

marshmallow_dataclass/__init__.py
```diff
@@ -34,8 +34,9 @@ class User:
       })
       Schema: ClassVar[Type[Schema]] = Schema # For the type checker
 """
-
+from enum import Enum, EnumMeta
 import dataclasses
+import inspect
```
lovasoa: Do we still need `inspect` ?

evanfwelch: I think the diffs are confusing because I inadvertently moved this import up a few lines. It's actually still used elsewhere in the `class_schema` function. 

Do you have a preferred order for imports? My instinct is normally a) python standard library, b) 3rd party packages, c) imports from within the package itself. Happy to change this or leave it alone!

lovasoa: Oh, you are right, sorry !

lovasoa:
It's marges; thank you again @evanfwelch !

PR #19, opened 2019-05-21
Fix default value handling for `NewType`s

karlb:


marshmallow_dataclass/__init__.py
```diff
@@ -270,6 +270,9 @@ def field_for_schema(
 
     >>> field_for_schema(NewType('UserId', int)).__class__
     <class 'marshmallow.fields.Integer'>
+
+    >>> field_for_schema(NewType('UserId', int), default=0).__class__
+    <class 'marshmallow.fields.Integer'>
```
lovasoa: Maybe we should rather add a test that explicitly tests for the bug being fixed here, by asserting that the default value was taken into account. Something like

```py
>>> field_for_schema(NewType('UserId', int), default=0).default
0
```

karlb: That's clearly a better test! I stopped thinking about the test when I got it to reproduce the unintended exception.

PR #25, opened 2019-07-29
Add support for union fields using marshmallow_union package

### LINKED ISSUE #24: Add Support for Unions
I absolutely love your library, it's made my life so much easier. Thank you!

Can we add support for using [marshmallow_union](https://python-marshmallow-union.readthedocs.io/en/latest/usage.html) on Union fields? 

Similar to how you support `marshmallow_enum`?

alextremblay:
Fixes #24 

This is a proposal to add support for Union types.

It has been tested as follows:

EDIT: fixed tests to ACTUALLY test marshmallow serialization
```python
from marshmallow import ValidationError
from typing import Union

@dataclass
class Test:
    union: Union[str, int]

t = Test('hello')
print(t.Schema._declared_fields['union'].__class__)  # <class 'marshmallow_union.Union'>
t2 = Test.Schema().load({'union': 'hello'})
t3 = Test.Schema().load({'union': 5})
try:
    t4 = Test.Schema().load({'union': False})
except ValidationError:
    pass
```

Thoughts?

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -270,7 +270,6 @@ def field_for_schema(
     >>> field_for_schema(Enum("X", "a b c")).__class__
     <class 'marshmallow_enum.EnumField'>
 
-    >>> import marshmallow_union
     >>> import typing
     >>> field_for_schema(typing.Union[int,str]).__class__
     <class 'marshmallow_union.Union'>
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -271,7 +271,8 @@ def field_for_schema(
     <class 'marshmallow_enum.EnumField'>
 
     >>> import marshmallow_union
-    >>> field_for_schema(Union[int,str]).__class__
+    >>> import typing
+    >>> field_for_schema(typing.Union[int,str]).__class__
     <class 'marshmallow_union.Union'>
 
     >>> field_for_schema(NewType('UserId', int)).__class__
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -270,6 +270,9 @@ def field_for_schema(
     >>> field_for_schema(Enum("X", "a b c")).__class__
     <class 'marshmallow_enum.EnumField'>
 
+    >>> field_for_schema(Union[int,str]).__class__
+    <class 'marshmallow_union.Union'>
+
     >>> field_for_schema(NewType('UserId', int)).__class__
     <class 'marshmallow.fields.Integer'>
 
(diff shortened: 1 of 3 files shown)


marshmallow_dataclass/__init__.py
```diff
@@ -329,6 +332,12 @@ def field_for_schema(
             metadata['missing'] = metadata.get('missing', None)
             metadata['required'] = False
             return field_for_schema(subtyp, metadata=metadata)
+        elif typing_inspect.is_union_type(typ):
+            subfields = []
+            for subtyp in arguments:
```
lovasoa: This may be more readable as a list comprehension.

lovasoa:
Hello, and thank you for your pull request ! This is a useful feature, and I am willing to merge this.

However, as I've told before in other issues, the feature would have to be implemented twice, in two pull requests: one on `master`, the other on `marshmallow3`. 

I you find this annoying, we can also simply wait for marshmallow to publish the first stable version of marshmallow 3. Several other feature requests are on hold until marshmallow 3 is released.

marshmallow_dataclass/__init__.py
```diff
@@ -270,6 +270,9 @@ def field_for_schema(
     >>> field_for_schema(Enum("X", "a b c")).__class__
     <class 'marshmallow_enum.EnumField'>
 
+    >>> field_for_schema(Union[int,str]).__class__
```
lovasoa: You need to import `Union` if you want to use it.

alextremblay:
Ok, I've made the changes you suggested. 

Unfortunately, I can't create a marshmallow2 version of this feature, because the `marshmallow_union` library requires marshmallow>=3.0.0rc6...

How do you want to proceed?

lovasoa:
Oh, and I think you need to update `.travis.yml` to have the tests pass:
https://github.com/lovasoa/marshmallow_dataclass/blob/master/.travis.yml#L12

lovasoa:
> Unfortunately, I can't create a marshmallow2 version of this feature, because the `marshmallow_union` library requires marshmallow>=3.0.0rc6...

OK, let's just keep it like that, then. I'll merge this.

marshmallow_dataclass/__init__.py
```diff
@@ -270,6 +270,10 @@ def field_for_schema(
     >>> field_for_schema(Enum("X", "a b c")).__class__
     <class 'marshmallow_enum.EnumField'>
 
+    >>> import marshmallow_union
```
lovasoa: You don't need to import `marshmallow_union` here.

marshmallow_dataclass/__init__.py
lovasoa: @alextremblay : You have to import [Union](https://docs.python.org/3/library/typing.html#typing.Union) from typing.

alextremblay: fixed :)

marshmallow_dataclass/__init__.py
alextremblay: fixed :)

alextremblay:
Sweet! Thank you :)

PR #28, opened 2019-09-13
Custom BaseSchema to derive from

manmime:
https://github.com/lovasoa/marshmallow_dataclass/issues/26

PR #35, opened 2019-09-18
Use marshmallow's built-in type mapping

sloria:


lovasoa:
nice, I didn't know marshmallow had a built in type mapping

lovasoa:
However, this PR would not be compatible with https://github.com/lovasoa/marshmallow_dataclass/pull/28
We would have to read the TYPE_MAPPING from the custom BaseSchema

sloria:
#28 does not respect the custom `BaseSchema`'s `TYPE_MAPPING`, so this isn't necessarily incompatible. It would be nice to implement that, but I think it's not strictly necessary before merging this PR or #28.

lovasoa:
Oh, @sloria I did not realize you were the main contributor of marshmallow. You must be much more knowledgeable about marshmallow than I am.
Would you like to be added as a collaborator to this project ? I am currently the only one, and I'm sure having a marshmallow contributor here would help a lot.

sloria:
I'd be happy to collaborate; we're using it work now so I have an interest in helping.

sloria:
By the way, if you'd like to join the marshmallow Slack channel, email me at sloria1 AT gmail.com and I'll invite you.

PR #31, opened 2019-09-18
Format with black

sloria:
Adds pre-commit configuration.

(Optionally) install pre-commit hooks with

```
pre-commit install
```

To run formatting manually:

```
pre-commit run --all-files
```

PR #36, opened 2019-09-19
Speed up travis builds

sloria:
Only install needed dependencies and build docs in deploy stage


sloria:
Is it necessary to test against all the point versions of 3.7?

sloria:
Just going to roll this into #31 because they're related

PR #37, opened 2019-09-19
Only test 3.7.0 and newest 3.7 stable

sloria:
Testing against 5 different py37 versions is overkill IMO


lovasoa:
I'm ok with that. I was just a little scared at the beginning, because we use private APIs from python's standard library, that could theoretically be broken even in a minor python version. But it never actually happened.

PR #38, opened 2019-09-19
Add pypy to the list of CI-tested versions

lovasoa:
Do not merge yet !

Hey @sloria ! I noticed that your recent addition of `black` breaks CI if we want to test with pypy. All the other tests pass with pypy.

Black's incompatibility with pypy is known both by pypy and black:
 - https://bitbucket.org/pypy/pypy/issues/2985/pypy36-osreplace-pathlike-typeerror
 - https://github.com/psf/black/issues/780

I feel like the advantages of testing with pypy are larger than those of using black.
What do you think about using a different formatter ?
We could also disable formatting tests with pypy, but since we just started adding formatting to CI, I feel it's still easy to change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "3.7.0"
   - "3.7"
   - "3.8-dev"
+  - "pypy3"
 install:
   - pipenv install --pre -e '.[dev]'
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,29 @@
 language: python
 dist: xenial
-python:
-  - "3.6"
-  - "3.7.0"
-  - "3.7"
-  - "3.8-dev"
-  - "pypy3"
+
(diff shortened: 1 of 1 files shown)


sloria:
> I feel like the advantages of testing with pypy are larger than those of using black.

I disagree. black has been a huge time-saver and bikeshed-painter. It also has wide adoption and reasonable output compared to other formatters.

I ended up removing pypy builds for my projects because 1) they increased CI build times, 2) I had trouble getting pypy installed locally when I wanted to reproduce a build failure, and 3) when a pypy build failure did occur, it was never because of my code (usually an issue on Travis's end).

Pypy is meant to be on parity--bug-to-bug-- with Cpython, so testing against it seems redundant. It'd be testing the language more than your code.

tl;dr--for pure-python projects, pypy builds are unnecessary and only add maintenance burden.

PR #40, opened 2019-09-20
Update documentation in README

sloria:


sloria:
The docs in the package should also be updated for consistency. I can do that in a follow-up PR.

sloria:
To see what this looks like: https://github.com/lovasoa/marshmallow_dataclass/tree/docs

lovasoa:
The `Building` example is a little verbose. It's nice to have it once in the beginning to expose what the library can do, but I would prefer to avoid repeating it several times. We can use the `Person` or `Point` example in the **`@dataclass` shortcut** and **Customizing generated fields** sections. What do you think ?

sloria:
Sounds good to me. I'm off to work on other things now; feel free to commit to my branch and revise as you see fit.

README.md
```diff
@@ -69,7 +69,9 @@ Install `marshmallow_dataclass<6.0` if you need marshmallow 2 compatibility.
 
 ## Usage
 
-Use the [`class_schema`](https://lovasoa.github.io/marshmallow_dataclass/html/marshmallow_dataclass.html#marshmallow_dataclass.class_schema) function to generate a marshmallow [Schema](https://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.Schema) class from a [`dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass).
+Use the [`class_schema`](https://lovasoa.github.io/marshmallow_dataclass/html/marshmallow_dataclass.html#marshmallow_dataclass.class_schema)
+function to generate a marshmallow [Schema](https://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.Schema)
```
sloria: Woops, I've been experimenting with [sentence-per-line-break](https://twitter.com/garybernhardt/status/1167123539834523648?s=20), so that's why all the long lines. Happy with whatever though.

README.md
```diff
@@ -89,58 +91,46 @@ PersonSchema = marshmallow_dataclass.class_schema(Person)
...
-    name: str = field(default="anonymous")
-
-
-@dataclass
-class City:
-    name: Optional[str]
-    buildings: List[Building] = field(default_factory=list)
+class Person:
+    name: str = field(
+        metadata=dict(description="The person's first name", load_only=True)
```
sloria: 👍

lovasoa:
@sloria, you can merge this if you have nothing to add

sloria:
@lovasoa Do you want to go ahead and rename the repo to marshmallow-dataclass? GitHub will take care of the redirects just fine. `setup.py` should also be updated.

sloria:
Also, what do you think about getting rid of the generated RST docs? It seems redundant to have docs in two places.

We can keep the docstrings--just proposing to remove the `docs/` directory.

lovasoa:
I don't think it's inconvenient to have the repo named marshmallow_dataclass. I'm not sure I understand what you mean about the duplicated docs... What I care about is the generated html documentation that is automatically generated and made available on github.io. I did not know about any rst docs.

lovasoa:
the rst files in the docs folder simply serve as templates when generating the html documentation for https://lovasoa.github.io/marshmallow_dataclass/html/marshmallow_dataclass.html

sloria:
I realize `docs/` just contains the templates. As a user, I think it's a little confusing to be pointed to the GitHub Pages page as the documentation when the narrative docs actually live in the README. Maybe we can just point out in the README that the GH Pages points to the "API Documentation".

sloria:
> I don't think it's inconvenient to have the repo named marshmallow_dataclass.

Yeah, it's super minor. Packaging tools like `pip-tools` normalize the `_` to a `-`, so it's sliiiightly nicer to use the dashes to avoid confusion.

Issue #42, opened 2019-10-06
Dataclasses do not implement __iter__, causing webargs.use_kwargs to fail


CptSpaceToaster:
webargs will attempt to combine the results of schema.load into the resulting kwargs to be passed into a function, as seen here:

https://github.com/marshmallow-code/webargs/blob/5.5.1/src/webargs/core.py#L446-L448

In order to repair this assumption, while still using dataclasses, I had to implement a mixin for my dataclasses to allow them to appear as iterables for the purposes of `kwargs.update(dataclass)`

```python
class Iterable:
    def __iter__(self):
        """When extending a dataclass, yields all key value pairs as if it were a dictionary"""
        for attr, value in self.__dict__.items():
            if value is not None:
                yield attr, value

@dataclass
class Building(Iterable):
    # field metadata is used to instantiate the marshmallow field
    height: float = field(metadata={"validate": marshmallow.validate.Range(min=0)})
    name: str = field(default="anonymous")
```

Should I go make noise over in webargs saying they should accept dataclasses? or is this something that we can safely hack in here, as we're are actively transforming the resulting dictionary INTO a dataclass.

Another acceptable solution would be to add a meta argument to allow a developer to skip the `post_load` step, and allow `load` to return a dictionary and NOT a dataclass, which is honestly more optimal IMO.

lovasoa:
The whole point of this library is to let you use typed objects instead of dicts. If you convert the result back to a dict and use that, you lose a lot of the advantages. I would open a feature request in webargs...

sloria:
I don't think it's something we'll add in webargs either (maintainer speaking). It's not intuitive for object attributes to be transformed to kwargs.

@CptSpaceToaster  I think your best bet would be to use a custom base schema class that overrides `make_data_class`, as suggested in https://github.com/lovasoa/marshmallow_dataclass/pull/43#issuecomment-538723437

lovasoa:
@sloria : I agree that transforming the object back into a dict does not make much sense, but what would you think about supporting schemas with a custom `post_load` in webargs ?

That would allow nice, readable, and typed checked web server code, like:

```py

@dataclass
class Person:
  name: str

@app.route("/")
@use_args(Person.Schema)
def index(person):
    return "Hello " + person.name
``` 

Can I open a feature request on [marshmallow-code/webargs](https://github.com/marshmallow-code/webargs) about this ?

sloria:
[...]

PR #43, opened 2019-10-06
Allow schema.load to return a dictionary

CptSpaceToaster:
Mentioned in https://github.com/lovasoa/marshmallow_dataclass/issues/42, this PR adds a meta argument that allows a developer to skip the post_load step, and allow `load` to return a dictionary and NOT a dataclass

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -455,7 +455,9 @@ def _base_schema(
         @marshmallow.post_load
         def make_data_class(self, data, **_):
             if skip_post_load:
-                return data
+                return {
+                    key: value for (key, value) in data.items() if value is not None
+                }
             return clazz(**data)
 
     return BaseSchema
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -154,7 +154,9 @@ def add_schema(_cls=None, base_schema=None):
 
 
 def class_schema(
-    clazz: type, base_schema: Optional[Type[marshmallow.Schema]] = None
+    clazz: type,
+    base_schema: Optional[Type[marshmallow.Schema]] = None,
+    skip_post_load=False,
 ) -> Type[marshmallow.Schema]:
 
     """
@@ -283,7 +285,11 @@ def class_schema(
         if field.init
     )
 
-    schema_class = type(clazz.__name__, (_base_schema(clazz, base_schema),), attributes)
+    schema_class = type(
+        clazz.__name__,
+        (_base_schema(clazz, base_schema, skip_post_load=skip_post_load),),
+        attributes,
+    )
     return cast(Type[marshmallow.Schema], schema_class)
 
 
@@ -437,7 +443,7 @@ def field_for_schema(
 
 
 def _base_schema(
-    clazz: type, base_schema: Type[marshmallow.Schema] = None
+    clazz: type, base_schema: Type[marshmallow.Schema] = None, skip_post_load=False
 ) -> Type[marshmallow.Schema]:
     """
     Base schema factory that creates a schema for `clazz` derived either from `base_schema`
@@ -448,6 +454,8 @@ def _base_schema(
     class BaseSchema(base_schema or marshmallow.Schema):  # type: ignore
         @marshmallow.post_load
         def make_data_class(self, data, **_):
+            if skip_post_load:
+                return data
             return clazz(**data)
 
     return BaseSchema


marshmallow_dataclass/__init__.py
```diff
@@ -448,6 +454,8 @@ def _base_schema(
     class BaseSchema(base_schema or marshmallow.Schema):  # type: ignore
         @marshmallow.post_load
         def make_data_class(self, data, **_):
+            if skip_post_load:
```
sloria: Can you even prevent `make_data_class` from being bound to the class?

```python
    if not skip_post_load:
        @marshmallow.post_load
        def make_data_class(....):
```

sloria:
The basic proposal here seems reasonable. Any objections @lovasoa?

lovasoa:
What I would prefer would be to use the post_load method from the custom base Schema if it's defined. This seems like a more natural and less surprising way to do things, and avoids changing the API for this niche use case.

sloria:
@lovasoa Fair enough. What do you think about changing the name of `make_data_class` to `make_object`? I think it's a more appropriate name and is more natural esp. when users override it to return a non-dataclass type like a dict.

CptSpaceToaster:
> What I would prefer would be to use the post_load method from the custom base Schema if it's defined

Are you proposing that developers should be able to supply their own `post_load` in place of the one being used? Or that if `skip_post_load=True`, that the factory should fall back upon some _default post load operation_ that I'm unaware of?  

If the user provides their own `post_load` operation either to the factory, or in their Schema, we could avoid adding the post_load operation to the chain that builds the dataclass. But then we've removed the ability to perform post_load operations on dataclasses, which for those of us that actually need dataclasses out of this thing... I think that's a big misstep.  I know I'd be satisfied, as I only want dictionaries out of `load()`, but I think removing that for others is poor.

> ... when users override it to return a non-dataclass type like a dict.

I'm must not be tracking. I can't find a way I can supply a custom BaseSchema from outside the module that will be consumed by the `_base_schema` factory. Are you suggesting we remove the `skip_post_load` and instead supply the class we're SUPPOSED to use with the factory and bubble that down through class_schema?

What I'm writing today:
```python
HealthcheckResponseSchema = class_schema(HealthcheckResponse, skip_post_load=True)
```

We could consider the following

```python
from marshmallow_dataclass import BaseSchema

class MyBaseSchema(BaseSchema):  # type: ignore
    def make_data_class(self, data, **_):
        # do nothing
        return data

HealthcheckResponseSchema = class_schema(HealthcheckResponse, base=MyBaseSchema)
```

However as `class BaseSchema` is currently being defined in a `_base_schema` factory, I'm unsure such changes will be easy.

marshmallow_dataclass/__init__.py
CptSpaceToaster: black wanted me to put an empty line after `if not skip_post_load:` for some reason?

```python
if not skip_post_load:

        @marshmallow.post_load
        def make_data_class(....):
```

CptSpaceToaster: Checks came in...... I'm not sure this works.... I thought it did, but travis isn't happy, and I can't grok why: https://travis-ci.org/lovasoa/marshmallow_dataclass/jobs/594249180

sloria:
The latest release (6.1.0) allows you to pass `base_schema` https://github.com/lovasoa/marshmallow_dataclass#customizing-the-base-schema

```python
HealthcheckResponseSchema = class_schema(HealthcheckResponse, base_schema=MyBaseSchema)
```

---

@lovasoa Related note: wdyt about doing more frequent stable version releases and marking the relevant API e.g. `NewType` as provisional? I think that would reduce a lot of confusion and increase adoption of recent releases (by and large, users don't install pre-releases).

CptSpaceToaster:
Ok, spending a little time with the project is getting me up to speed here. Thanks for being patient.

https://github.com/lovasoa/marshmallow_dataclass/blob/v6.1.0rc2/marshmallow_dataclass/__init__.py#L448-L451

I guess I'm objecting to having to have to either add an Iterable mixing, or being forced to implement a BaseSchema to do this... 
```python
import marshmallow

class MyBaseSchema(marshmallow.Schema):
    @marshmallow.post_load
    def make_data_class(self, data, **_):
        return {key:value for (key, value) in data.items() if value is not None}

HealthcheckResponseSchema = class_schema(HealthcheckResponse, base_schema=MyBaseSchema)
```

Especially because while I like being able to get Marshmallow dataclasses out of this... I never really asked for them. I really just wanted a way to create nested schemas via dataclasses... but I need the end-result to produce dictionaries to work with webargs...

lovasoa:
@CptSpaceToaster, so you actually never use the classes you write, only the schemas you generate from them ? What exactly do you need `marshmallow_dataclass` for, then ?

CptSpaceToaster:
I tend to use dataclasses I write to generate response objects, before shoving the resulting object into a `Schema.dump(my_dataclass)`... meaning I tend to only rely on them when I generate output.

```python
def get(self):
    # Interact with Database, or do work
    response = MyDataclass(
        foo=bar,
        ...
    )
    return HealthcheckResponseSchema().dump(response), 200
```

I was attempting to also rely on the schemas for input similar to your [comment](https://github.com/lovasoa/marshmallow_dataclass/issues/42#issuecomment-538888118), and webargs throws a fit unless support is added here to meet its expectations, or its expectations are changed.

While I can go back to vanilla marshmallow to generate most of what I need, marshmallow_dataclass moves my project into the realm of dataclasses as a whole (which is a nice bonus) while dataclasses themselves are (IMO) easier to create, maintain, and nest!

```python
@dataclass
class Task:
    description: str = None
    title: str

@dataclass
class TaskList:
    tasks: List[Task]
    date: Date

TaskSchema = class_schema(Task)
TaskListSchema = class_schema(TaskList)
```

Instead of the vanilla equivalent

```python
class TaskSchema(Schema):
    description = fields.Str()
    title = fields.Str(required=True)

class TaskListSchema(Schema):
    tasks = fields.Nested(TaskSchema, required=True)
    last_updated = fields.Date(required=True)
```

sloria:
Can you use `use_args` instead of `use_kwargs`? That _should_ just work for schemas that have a post_load method.

CptSpaceToaster:
[Yep!](https://github.com/lovasoa/marshmallow_dataclass/issues/42#issuecomment-539057029)

Sorry for the trouble, thanks for being patient.

Sounds like something that could be added to the documentation!

sloria:
Great. Feel free to send a docs PR on webargs =)

Closing this for now

Issue #47, opened 2019-10-25
Error when dataclass member names conflicting with Schema members


dutekvejin:
Hi,

I'm hitting the issue when my dataclass defines property decorated method that is named `context`

Here is an example:

```python
from dataclasses import dataclass

from marshmallow_dataclass import class_schema


@dataclass
class MyClazz:
    foo: str

    @property
    def context(self) -> None:
        return None


MyClazzSchema = class_schema(MyClazz)
MyClazzSchema().load({'foo': 'bar'})

```

and the result:

```
Traceback (most recent call last):
  File "/scratches/issue.py", line 16, in <module>
    MyClazzSchema().load({'foo': 'bar'})
  File "/scratches/venv/lib/python3.7/site-packages/marshmallow/schema.py", line 388, in __init__
    self.context = context or {}
AttributeError: can't set attribute
```

And this is happening with any dataclass member that is conflicting with `marshmallow.Schema` attribute names eg. `many`,  `load_only`, `fields`... 

When I chased down this issue, I found that `marshmallow_dataclass.class_schema` copying all public members of the dataclass to the schema. 
https://github.com/lovasoa/marshmallow_dataclass/blob/445a0efc4b93f58b178cca3670809768b098c4a7/marshmallow_dataclass/__init__.py#L272-L273

When I change that particular line of code to `attributes = {}` everything starts to work fine as `marshmallow_dataclass.class_schema` will create `marshmallow.fields.Field` from every dataclass declared field and pass it into created schema.

For me the solution would be  not to add all public members of the dataclass to the schema but just dataclass declared field. I can create PR with this change but I'm not sure if this can break something, and I'm wondering why copying all public members at the first place?

Thanks

dutekvejin:
@lovasoa that can be achieved by declaring **internal** whitelist of member names that can be passed through, then `class_schema` can extract whitelisted members from given dataclass.

If you are ok with this, I'll create PR in next few days?

lovasoa:
Yes, please submit a PR :) Don't forget to add tests.

PR #48, opened 2019-10-25
Filter out non required members from schema attributes

dutekvejin:
#47 

Allowing just whitelisted members was not enough, so I had to add one more filter 
 to allow marshmallow hooks to be declared and copied to schema. I also added tests to check if those filters are applied.

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -273,8 +273,9 @@ def class_schema(
                 f"{getattr(clazz, '__name__', repr(clazz))} is not a dataclass and cannot be turned into one."
             )
 
-    # Copy all whitelisted members of the dataclass to the schema.
-    attributes = {k: v for k, v in inspect.getmembers(clazz) if k in MEMBERS_WHITELIST}
+    # Copy all marshmallow hooks and whitelisted members of the dataclass to the schema.
+    attributes = {k: v for k, v in inspect.getmembers(clazz)
+                  if hasattr(v, '__marshmallow_hook__') or k in MEMBERS_WHITELIST}
     # Update the schema members to contain marshmallow fields instead of dataclass fields
     attributes.update(
         (
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -61,7 +61,7 @@ __all__ = ["dataclass", "add_schema", "class_schema", "field_for_schema", "NewTy
 NoneType = type(None)
 _U = TypeVar("_U")
 
-# Whitlist of dataclass members that will be copyied to generated schema.
+# Whitelist of dataclass members that will be copied to generated schema.
 MEMBERS_WHITELIST: Set[str] = {"Meta"}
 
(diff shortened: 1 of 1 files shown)


lovasoa:
You can run the pre-commit hooks on your computer to make sure you commit only properly formatted code.

marshmallow_dataclass/__init__.py
```diff
@@ -60,6 +61,9 @@ class User:
 NoneType = type(None)
 _U = TypeVar("_U")
 
+# Whitlist of dataclass members that will be copyied to generated schema.
```
sloria:
```suggestion
# Whitelist of dataclass members that will be copied to generated schema.
```

marshmallow_dataclass/__init__.py
```diff
@@ -256,6 +260,53 @@ def class_schema(
...
+    ...     @marshmallow.post_load
+    ...     def post_load(self, *args, **kwargs):
+    ...         pass
+    ...     def custom_method(self, *args, **kwargs):
+    ...         pass
+    ...     @property
+    ...     def custom_property(self, *args, **kwargs):
+    ...         return None
+
+    >>> hasattr(class_schema(Anything), 'Meta')
```
sloria: Nitpick: maybe store `class_schema(Anything)` in a variable to DRY things a bit.

marshmallow_dataclass/__init__.py
```diff
@@ -256,6 +260,53 @@ def class_schema(
...
+    ...         pass
+    ...     @marshmallow.post_dump
+    ...     def post_dump(self, *args, **kwargs):
+    ...         pass
+    ...     @marshmallow.pre_load
+    ...     def pre_load(self, *args, **kwargs):
+    ...         pass
+    ...     @marshmallow.post_load
+    ...     def post_load(self, *args, **kwargs):
+    ...         pass
```
lovasoa: Maybe just one or two marshmallow hooks would have been enough, in order to keep the doctests small ?

sloria:
I guess it could be convenient to set hooks on the dataclass without having to declare a separate base schema class. I don't have a strong opinion either way.

@lovasoa Feel free to merge when you see fit.

lovasoa:
@sloria If I'm understanding it correctly, this PR still allows setting hook on the dataclass.

sloria:
@lovasoa Yeah, in my previous comment I was suggesting *not* supporting hooks on the dataclass. But I think it's fine to keep.

dutekvejin:
@sloria if you ask me :), I agree that dataclass should not define any hook but rather use `base_schema`. i just let that behaviour unchanged as  I didn't want to break BC more than it's needed.

marshmallow_dataclass/__init__.py
dutekvejin: No nitpicking :)

marshmallow_dataclass/__init__.py
dutekvejin: Done

PR #58, opened 2019-12-13
 Test the mypy plugin on windows

### LINKED ISSUE #57: Test the mypy plugin on windows
Since @selimb added support for mypy plugin tests on windows in https://github.com/typeddjango/pytest-mypy-plugins/pull/14 , we'll be able to remove this condition in setup.py as soon as a new version of pytest-mypy-plugins comes out : 

https://github.com/lovasoa/marshmallow_dataclass/blob/master/setup.py#L27

lovasoa:
Fixes #57

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,8 +22,6 @@ EXTRAS_REQUIRE = {
         "pytest",
         # re: pypy: typed-ast (a dependency of mypy) fails to install on pypy
         # https://github.com/python/typed_ast/issues/111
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -24,7 +24,7 @@
         # https://github.com/python/typed_ast/issues/111
         # re: win32: pytest-mypy-plugins depends on capturer, which isn't supported on
```
sloria: This comment should be removed.

PR #59, opened 2019-12-14
Exclude tests directory from published package

tomskikh:
For now published package contains a top-level module "tests". It's causing to fail tests on dependent projects with "cannot import module tests.foo.bar" error.


lovasoa:
Thank you for your contribution. I am going to publish 7.2.1.

sloria:
FWIW, the proper way to include `tests` appears to be `graft tests` in `MANIFEST.in`: https://github.com/marshmallow-code/webargs/pull/446/files

PR #69, opened 2020-02-15
switch 'type is' to 'isinstance' for enum check

kristian-lindin:
When using an Enum that has extended the EnumMeta (Djangos 'TextChoices' is a concrete example), the 'type(typ) is EnumMeta' fails the "Enum type check". Changing this to 'isinstance(typ, EnumMeta)' will pass the "Enum type check" also for these enums with an extended EnumMeta.


lovasoa:
Looks good to me. Thank you !

PR #70, opened 2020-02-20
Add basic cache support

ZoranPavlovic:
When dynamically creating marshmallow schemas and registering them with marshmallow registry (for ApiSpec using the marshmallow plugin). This is an attempt at returning the same "schema" instance when the same input type is used to generate it.

Otherwise, marshmallow sees them as completely unique schemas despite having the same name + fields.

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -157,7 +157,7 @@ def add_schema(_cls=None, base_schema=None):
     return decorator(_cls) if _cls else decorator
 
 
-SCHEMA_REGISTRY: Dict[type, Type[marshmallow.Schema]] = {}
+SCHEMA_REGISTRY: Dict[Tuple[type, Optional[Type[marshmallow.Schema]]], Type[marshmallow.Schema]] = {}
 
 
 def class_schema(
@@ -274,8 +274,30 @@ def class_schema(
     Traceback (most recent call last):
     ...
     marshmallow.exceptions.ValidationError: {'name': ['Name too long']}
+    >>> @dataclasses.dataclass
+    ... class Simple:
+    ...     one: str = dataclasses.field()
+    ...     two: str = dataclasses.field()
+    >>> @dataclasses.dataclass
+    ... class ComplexNested:
+    ...     three: int = dataclasses.field()
+    ...     four: Simple = dataclasses.field()
+    >>> id(class_schema(ComplexNested)) == id(class_schema(ComplexNested))
+    True
+    >>> class_schema(ComplexNested) is class_schema(ComplexNested)
+    True
+    >>> id(class_schema(Simple)) == id(class_schema(Simple))
+    True
+    >>> class_schema(Simple) is class_schema(Simple)
+    True
+    >>> class_schema(Simple) is class_schema(ComplexNested)._declared_fields["four"].nested
+    True
+    >>> len(set([class_schema(ComplexNested), class_schema(ComplexNested, base_schema=None), class_schema(ComplexNested, None)]))
+    1
+    >>> len(set([class_schema(Simple), class_schema(Simple, base_schema=None), class_schema(Simple, None)]))
+    1
     """
-    cached_schema = SCHEMA_REGISTRY.get(clazz, None)
+    cached_schema = SCHEMA_REGISTRY.get((clazz, base_schema), None)
     if cached_schema is not None:
         return cached_schema
 
@@ -310,7 +332,7 @@ def class_schema(
 
     schema_class = type(clazz.__name__, (_base_schema(clazz, base_schema),), attributes)
     result = cast(Type[marshmallow.Schema], schema_class)
-    SCHEMA_REGISTRY[clazz] = result
+    SCHEMA_REGISTRY[(clazz, base_schema)] = result
     return result
 
 
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -68,6 +68,7 @@ MEMBERS_WHITELIST: Set[str] = {"Meta"}
 # Max number of generated schemas that class_schema keeps of generated schemas. Removes duplicates.
 MAX_CLASS_SCHEMA_CACHE_SIZE = 1024
 
+
 # _cls should never be specified by keyword, so start it with an
 # underscore.  The presence of _cls is used to detect if this
 # decorator is being called with parameters or not.
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -313,8 +313,7 @@ def _proxied_class_schema(
     )
 
     schema_class = type(clazz.__name__, (_base_schema(clazz, base_schema),), attributes)
-    result = cast(Type[marshmallow.Schema], schema_class)
-    return result
+    return cast(Type[marshmallow.Schema], schema_class)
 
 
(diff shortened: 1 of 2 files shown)


ZoranPavlovic:
@lovasoa - Let me know if this is the direction you're happy with, or if you'd like it implemented some other way?

lovasoa:
Maybe `functools.lru_cache` from the standard library could be used instead of rolling our own cache implementation ?

ZoranPavlovic:
@lovasoa - I don't mind doing that. However, the lru_cache is a memoization decorator that "invalidates" the cache if any of the params to the function change. Currently there are two params in that function: the dataclass and base_schema.

Would you say that a different base_schema should yield a different/duplicate instance, or should it return the original regardless of the different base_schema passed through? I don't know too much about the internals in that regard and would appreciate your input.

lovasoa:
Of course, the function should always return a Schema that matches the given parameters. If you call the function with a certain base_schema, and the function returns a schema with a different base_schema, then it's a bug.

ZoranPavlovic:
@lovasoa - Thanks will do

However, don't think lru_cache will work for this. Base_schema is an optional and as such causes a cache miss even though the value is the same. E.g. the following code fails:

```python
import dataclasses


@dataclasses.dataclass
class Simple:
    one: str = dataclasses.field()
    two: str = dataclasses.field()


schema_1a = class_schema(Simple) # base_schema is None here due to signature.
schema_1b = class_schema(Simple, base_schema=None)
schema_1c = class_schema(Simple, None)

print(id(schema_1a), id(schema_1b), id(schema_1c))  # 3 different instances
```

lovasoa:
I think it is acceptable in practice to return different instances when the function is invoked differently.
And if you really want to handle this case, I think you still shouldn't roll your own cache implementation. You can just create a proxy function without arguments on which you'll apply `lru_cache`.

marshmallow_dataclass/__init__.py
```diff
@@ -271,7 +274,32 @@ def class_schema(
...
+    >>> id(class_schema(Simple)) == id(class_schema(Simple))
+    True
+    >>> class_schema(Simple) is class_schema(Simple)
+    True
+    >>> class_schema(Simple) is class_schema(ComplexNested)._declared_fields["four"].nested
+    True
+    >>> len(set([class_schema(ComplexNested), class_schema(ComplexNested, base_schema=None), class_schema(ComplexNested, None)]))
+    1
+    >>> len(set([class_schema(Simple), class_schema(Simple, base_schema=None), class_schema(Simple, None)]))
+    1
```
lovasoa: Maybe this would be better in a separate test file than in the function's doc comment ?

marshmallow_dataclass/__init__.py
```diff
@@ -303,7 +331,9 @@ def class_schema(
     )
 
     schema_class = type(clazz.__name__, (_base_schema(clazz, base_schema),), attributes)
-    return cast(Type[marshmallow.Schema], schema_class)
+    result = cast(Type[marshmallow.Schema], schema_class)
+    SCHEMA_REGISTRY[(clazz, base_schema)] = result
```
lovasoa: Your cache can only grow. This is a memory leak.

ZoranPavlovic: I was going to add a max-size but then decided against it as that would re-introduce the duplicate issue, just at some arbitrary point further down. E.g. after 128 generated schemas.

Furthermore, it's not a cache per se, it's more of a registry. It'll grow to equal the size of unique dataclass/schemas and then stop growing.

marshmallow_dataclass/__init__.py
ZoranPavlovic: > I think it is acceptable in practice to return different instances when the function is invoked differently.

Yes, but in this instance it is unusual behavior and non-intuitive. E.g. giving the same dataclass and getting different schema instances back based on how you "called" the function.

> And if you really want to handle this case, I think you still shouldn't roll your own cache implementation. You can just create a proxy function without arguments on which you'll apply `lru_cache`.

It's a necessary case due to nested schemas. Two complex dataclasses that reference other custom dataclasses generate duplicates because the function to generate them gets called internally inside the marshmallow-dataclass functions. If I could have resolved this issue without making changes to the lib, I would have done so. If you have any suggestions as to how we can do that, that'll be awesome?

ZoranPavlovic: @lovasoa Will move the tests to a separate file.

marshmallow_dataclass/__init__.py
lovasoa: There is no guarantee that the number of unique schemas is bounded. I agree that this means we cannot **guarantee** that we will always return the same schema, but I think that's fine, or at least better than a memory leak.

ZoranPavlovic: Okay sure, fair point. What number would you suggest? 1024, perhaps? Default is 128 in lru_cache, which might cause issues more the general use-case.

tests/test_class_schema.py
```diff
@@ -0,0 +1,34 @@
...
+            one: str = dataclasses.field()
+            two: str = dataclasses.field()
+
+        @dataclasses.dataclass
+        class ComplexNested:
+            three: int = dataclasses.field()
+            four: Simple = dataclasses.field()
+
+        self.assertEqual(id(class_schema(ComplexNested)), id(class_schema(ComplexNested)))
+        self.assertTrue(class_schema(ComplexNested) is class_schema(ComplexNested))
```
lovasoa: Why not [`assertIs`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIs) ?

marshmallow_dataclass/__init__.py
lovasoa: Yes, 1024 seems reasonable 👍

ZoranPavlovic:
@lovasoa - All changes made. Lru_cache, 1024 max, using proxy method. Separate file for unit tests and applied black formatting.

marshmallow_dataclass/__init__.py
```diff
@@ -303,7 +313,8 @@ def class_schema(
     )
 
     schema_class = type(clazz.__name__, (_base_schema(clazz, base_schema),), attributes)
-    return cast(Type[marshmallow.Schema], schema_class)
+    result = cast(Type[marshmallow.Schema], schema_class)
+    return result
```
lovasoa: We can keep 
```py
 return cast(Type[marshmallow.Schema], schema_class)
```

lovasoa:
This looks good to me. Once the `return` is changed, I'll merge.

marshmallow_dataclass/__init__.py
ZoranPavlovic: Resolved

lovasoa:
Thank you for your contribution, @ZoranPavlovic !

ZoranPavlovic:
@lovasoa - Np, glad to help!

Btw, do you know if the changes have been merged in fully? It appears that they've made it into the master branch and a new (7.3) release has been made. But I'm not seeing the changes on the github release and on pypi?

lovasoa:
No, your commits are indeed not part of the latest release. You can always check that on https://github.com/lovasoa/marshmallow_dataclass/compare/v7.0.0...master 

I'll try to make a new release soon.

lovasoa:
It is now fixed. I deployed your code as 7.4.0. Sorry for the inconvenience, and thank you for letting me know.

ZoranPavlovic:
@lovasoa Np, thank you for the update and getting it in so quick :)

PR #76, opened 2020-04-19
Fix inconsistency in post_load behavior

### LINKED ISSUE #75: Behavior of marshmallow.post_load is dependent on function name
I discovered a discrepancy with how the `post_load` decorator behaves within a marshmallow dataclass. In [Schema._invoke_processors](https://github.com/marshmallow-code/marshmallow/blob/76196abf35ff9ec58f3dc2377ea7a8a9bf23712a/src/marshmallow/schema.py#L1192) the post loads `make_data_class` processor runs before or after the post_load defined in the dataclass based on whichever comes first alphabetically. Some example below illustrate this: 

```python
import marshmallow
import marshmallow_dataclass

@marshmallow_dataclass.dataclass
class Person:
	name: str

	@marshmallow.post_load
	def a(schema, data, **kwargs):
		data.name = data.name.capitalize()  # works if: data['name'] = data['name'].capitalize()
		return data

Person.Schema().load({'name': 'matt'})
>> AttributeError: 'dict' object has no attribute 'name'
```

```python
import marshmallow
import marshmallow_dataclass

@marshmallow_dataclass.dataclass
class Person:
	name: str

	@marshmallow.post_load
	def z(schema, data, **kwargs):
		data.name = data.name.capitalize()
		return data

Person.Schema().load({'name': 'matt'})
>> Person(name='Matt')
```

The post_load in the first example does not create an instance of the dataclass itself before being invoked and feels more so like a `pre_load` because of this. The validation checks still seem to work properly, but it just feels a little weird to work with. I know the documentation does not include anything about pre/post loading, but I figured I'd document this in case anyone else runs into the same issue.

lovasoa:
Fixes #75

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -35,7 +35,6 @@ Full example::
       Schema: ClassVar[Type[Schema]] = Schema # For the type checker
 """
 import inspect
-import typing
 from enum import EnumMeta
 from functools import lru_cache
 from typing import (
@@ -462,7 +461,7 @@ def _base_schema(
     # Remove `type: ignore` when mypy handles dynamic base classes
     # https://github.com/python/mypy/issues/2813
     class BaseSchema(base_schema or marshmallow.Schema):  # type: ignore
-        def load(self, data: typing.Mapping, *, many: bool = None, **kwargs):
+        def load(self, data: Mapping, *, many: bool = None, **kwargs):
             all_loaded = super().load(data, many=many, **kwargs)
             if many:
                 return [clazz(**loaded) for loaded in all_loaded]
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -34,10 +34,10 @@ Full example::
       })
       Schema: ClassVar[Type[Schema]] = Schema # For the type checker
 """
-import dataclasses
 import inspect
-from functools import lru_cache
+import typing
 from enum import EnumMeta
+from functools import lru_cache
 from typing import (
     overload,
     Dict,
@@ -54,6 +54,7 @@ from typing import (
     Set,
 )
 
+import dataclasses
 import marshmallow
 import typing_inspect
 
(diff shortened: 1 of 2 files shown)


marshmallow_dataclass/__init__.py
```diff
@@ -457,12 +457,16 @@ def _base_schema(
...
+
     # Remove `type: ignore` when mypy handles dynamic base classes
     # https://github.com/python/mypy/issues/2813
     class BaseSchema(base_schema or marshmallow.Schema):  # type: ignore
-        @marshmallow.post_load
-        def make_data_class(self, data, **_):
-            return clazz(**data)
+        def load(self, data: Mapping, *, many: bool = None, **kwargs):
+            all_loaded = super().load(data, many=many, **kwargs)
+            if many:
```
sloria, line 466: I think you'll need

```suggestion
            many = self.many if many is None else bool(many)
            if many:
```

to handle `many` passed to the schema ctor.

tests/test_post_load.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+
+import marshmallow
+
+import marshmallow_dataclass
+
+
+class TestFieldForSchema(unittest.TestCase):
```
sloria, line 8: It might be worth adding a test for `many=True`.

tests/test_post_load.py
```diff
@@ -0,0 +1,35 @@
...
+            @marshmallow.post_load
+            def a(self, data, **_kwargs):
+                data["name"] = data["name"].capitalize()
+                return data
+
+        actual = Person.Schema().load({"name": "matt"})
+        expected = Person(name="Matt")
+        self.assertEqual(actual, expected)
+
+    def test_postload_z(self):
```
sloria, line 23: My eyes might be failing me, but is there a difference between this test and `test_postload_a`?

lovasoa: The only difference is the name of the post_load function. See https://github.com/lovasoa/marshmallow_dataclass/issues/75

sloria: Oh, interesting--didn't think the post load functions would run in deterministic order. OK, it might make it clearer to have a single test with both schemas and sets of assertions and rename it to `test_post_load_method_naming_does_not_affect_data` or something like that. Could even add a code comment above it with

```python
# Regression test for https://github.com/lovasoa/marshmallow_dataclass/issues/75
```

so our future selves won't be confused by this.

No biggie, though--your call

tests/test_post_load.py
lovasoa: done

tests/test_post_load.py
lovasoa: done

lovasoa:
Thank you for your review @sloria ! I'm merging this.

PR #77, opened 2020-04-20
Improve typehint

heckad:


marshmallow_dataclass/__init__.py
```diff
@@ -161,9 +163,17 @@ def decorator(clazz: Type[_U]) -> Type[_U]:
...
-def class_schema(
-    clazz: type, base_schema: Optional[Type[marshmallow.Schema]] = None
-) -> Type[marshmallow.Schema]:
+@typing.overload
+def class_schema(clazz: type, base_schema: None) -> Type[marshmallow.Schema]:
+    ...
+
+
+@typing.overload
+def class_schema(clazz: type, base_schema: T_SchemaType) -> T_SchemaType:
```
lovasoa, line 172: Is this annotation correct though ? The object returned by this function is a class that inherits from T_SchemaType, not T_SchemaType itself...

lovasoa:
Can you give context about why this is needed ?

heckad:
```
import typing
from typing import Union

import marshmallow_dataclass
from marshmallow import Schema as BaseSchema
from marshmallow import SchemaOpts as BaseSchemaOpts
from marshmallow import types


class SchemaOpts(BaseSchemaOpts):
    def __init__(self, meta, **kwargs):
        super().__init__(meta, **kwargs)
        self.required_fields = getattr(meta, "required_fields", set())


class MySchema(BaseSchema):
    OPTIONS_CLASS = SchemaOpts

    def __init__(self, *, required_fields=set(), **kwargs):
        self.required_fields = required_fields or self.opts.required_fields
        super().__init__(**kwargs)

    def on_bind_field(self, field_name, field_obj):
        super().on_bind_field(field_name, field_obj)
        if field_name in self.required_fields:
            field_obj.required = True

    def make_data_class(self, data, **_):
        super().make_data_class(data, **_)

    def loads(
        self,
        json_data: Union[str, bytes, bytearray],
        *,
        many: bool = None,
        partial: typing.Union[bool, types.StrSequenceOrSet] = None,
        unknown: str = None,
        **kwargs
    ):
        return super().loads(
            json_data,  # type: ignore
            many=many, partial=partial, unknown=unknown, **kwargs)


@dataclass
class A:
    a: int


schema = marshmallow_dataclass.class_schema(A, base_schema=MySchema)
schema().loads(b'')
```
Mypy output
```
test.py:52: error: Argument 1 to "loads" of "Schema" has incompatible type "bytes"; expected "str"
Found 1 error in 1 file (checked 1 source file)
```
Thats because mypy thing that `schema` have type `Schema` instead of `MySchema`

marshmallow_dataclass/__init__.py
heckad: This function should return the same type that sanded throw `base_schema`. If `base_schema` is None then return default type.

lovasoa: Yes, I understand the problem with the two different overloads for this function. However, I don't think the type annotation you added is correct. The type of base_schema and the output type of the function are not the same.

lovasoa:
Also, can you fix the tests ? You can run `pre-commit` to check your work before pushing it.

heckad:
Done

heckad:
Thanks a lot!

lovasoa:
@heckad : I'm trying to write tests for this and it doesn't seem to be supported by mypy, though...

```
error: Unsupported type Type[T_Schema?]
```


https://github.com/python/mypy/issues/4300

lovasoa:
This makes me think we should probably wait for python/mypy#4300 (opened in 2017) to be fixed before merging this PR. The `Unsupported type Type[T_Schema?]` cannot be silenced, so I fear merging this may cause a lot of troubles to users...

lovasoa:
type hints should have been fixed

PR #81, opened 2020-05-02
Accept "None" as a value for fields of type "Any"

### LINKED ISSUE #80: Bug: not  allow null in any values
```py
from typing import Any
from typing import Dict

@dataclass()
class A:
    data: Dict[str, Any]


A.Schema().load({
    'data': {'inner_data': None}
})
```

lovasoa:
Fixes #80


marshmallow_dataclass/__init__.py
```diff
@@ -378,6 +375,9 @@ def field_for_schema(
     if field:
         return field(**metadata)
 
+    if typ is Any:
+        return marshmallow.fields.Raw(allow_none=True, **metadata)
```
sloria, line 379: Should this handle

```python
    my_field: Any = field(allow_none=False)
```

PR #84, opened 2020-05-11
mypy plugin support for custom dataclass decorator

spapinistarkware:
This change is [Reviewable](https://reviewable.io/reviews/lovasoa/marshmallow_dataclass/84)

### DIFF
--- a/marshmallow_dataclass/mypy.py
+++ b/marshmallow_dataclass/mypy.py
@@ -3,6 +3,7 @@ from typing import Callable, Optional, Type
 
 from mypy import nodes
 from mypy.plugin import DynamicClassDefContext, Plugin
+from mypy.plugins import dataclasses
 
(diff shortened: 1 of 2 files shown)


marshmallow_dataclass/mypy.py
```diff
@@ -21,6 +21,13 @@ def get_dynamic_class_hook(
             return new_type_hook
         return None
 
+    def get_class_decorator_hook(self, fullname: str):
+        from mypy.plugins import dataclasses
```
lovasoa, line 25: Shouldn't this be grouped with other imports ?

lovasoa:
Also, this breaks the tests. You can run `pre-commit` to automatically reformat the files and make the tests pass.

lovasoa:
thanks @spapinistarkware 🙏

PR #85, opened 2020-05-13
Allow to map a custom Field class to a Type

### LINKED ISSUE #66: Use NewType for derived List class
Hi there!

Coming from vanilla marshmallow, I had a `List` class that was deriving from `marshmallow.fields.List` and was adding some logic to how it was reading the input data on `_deserialize`.

Now, I am trying to integrate this with `marshmallow_dataclass` and hence create a new type. I have created a few other types that were not lists and it worked well, but I can't get it to work for lists. Here is what I have tried so far:

```python3
import typing

import marshmallow
from marshmallow_dataclass import dataclass, NewType


class MyList(marshmallow.fields.List):
    """A list field type that properly handles lists in MultiDict"""

    def _deserialize(self, value, attr, data, **kwargs):
        # removed the code, just checking if this is called
        raise Exception("Finally Called")


ListType = NewType("ListType", typing.List, field=MyList)


@dataclass 
class TestDataClass1:
    """
    This is the one I was expecting would work, as I thought that it would pass
    `int` to `MyList` which would come back to something similar to vanilla
    marshmallow i.e. `values = MyList(Int())`
    """
    values: ListType(int)

# fails with 'Not a valid integer.'
TestDataClass1.Schema().load({"values": [1,2,3]})

# Second try!
@dataclass 
class TestDataClass2: 
    values: ListType(typing.List[int])

# list is properly loaded, but that means MyList._deserialize was not called
TestDataClass2.Schema().load({"values": [1,2,3]})
```

I looked in the doc and example but could not find a reference for this use case, could you help me figure it out?

CedricCabessa:
This patch allow us to map a custom implementation of a field to map to a
generic type.
eg:

```
def dataclass(*args, **kwargs):
    kwargs["generic_types"] = kwargs.get("generic_types", {typing.List: MyList})
    return marshmallow_dataclass.dataclass(*args, **kwargs)

@dataclass
class MyClass:
  foo: List[int]
```

`foo` will be a `MyList` field and not a `marshmallow.fields.List`

Credits goes to @ybadiss to have developed this patch

Fixes #66

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -81,14 +81,12 @@ def dataclass(
     unsafe_hash: bool = False,
     frozen: bool = False,
     base_schema: Optional[Type[marshmallow.Schema]] = None,
-    generic_types: Optional[Dict[Type, Type[marshmallow.Schema]]] = None,
 ) -> Union[Type[_U], Callable[[Type[_U]], Type[_U]]]:
     """
     This decorator does the same as dataclasses.dataclass, but also applies :func:`add_schema`.
     It adds a `.Schema` attribute to the class object
 
     :param base_schema: marshmallow schema used as a base class when deriving dataclass schema
-    :param generic_types: dict to map generic type to custom marshmallow schema
 
     >>> @dataclass
     ... class Artist:
@@ -112,8 +110,8 @@ def dataclass(
         _cls, repr=repr, eq=eq, order=order, unsafe_hash=unsafe_hash, frozen=frozen
     )
     if _cls is None:
-        return lambda cls: add_schema(dc(cls), base_schema, generic_types)
-    return add_schema(dc, base_schema, generic_types)
+        return lambda cls: add_schema(dc(cls), base_schema)
+    return add_schema(dc, base_schema)
 
 
 @overload
@@ -124,28 +122,24 @@ def add_schema(_cls: Type[_U]) -> Type[_U]:
 @overload
 def add_schema(
     base_schema: Type[marshmallow.Schema] = None,
-    generic_types: Optional[Dict[Type, Type[marshmallow.Schema]]] = None,
 ) -> Callable[[Type[_U]], Type[_U]]:
     ...
 
 
 @overload
 def add_schema(
-    _cls: Type[_U],
-    base_schema: Type[marshmallow.Schema] = None,
-    generic_types: Optional[Dict[Type, Type[marshmallow.Schema]]] = None,
+    _cls: Type[_U], base_schema: Type[marshmallow.Schema] = None
 ) -> Type[_U]:
     ...
 
 
-def add_schema(_cls=None, base_schema=None, generic_types=None):
+def add_schema(_cls=None, base_schema=None):
     """
     This decorator adds a marshmallow schema as the 'Schema' attribute in a dataclass.
     It uses :func:`class_schema` internally.
 
     :param type cls: The dataclass to which a Schema should be added
     :param base_schema: marshmallow schema used as a base class when deriving dataclass schema
-    :param generic_types: dict to map generic type to custom marshmallow schema
 
     >>> class BaseSchema(marshmallow.Schema):
     ...   def on_bind_field(self, field_name, field_obj):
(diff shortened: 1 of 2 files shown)


YBadiss:
Very useful to us, and it fixes my issue #66 ! 👍

lovasoa:
Hello ! I like the idea of being able to handle custom serialization of collections. However, I don't really like the overhead the new parameter adds. Do you think we could use the existing `base_schema.TYPE_MAPPING` for this ?

marshmallow_dataclass/__init__.py
```diff
@@ -383,18 +403,34 @@ def field_for_schema(
...
+            **(base_schema.TYPE_MAPPING if base_schema else {}),
+        }
         if origin in (list, List):
-            child_type = field_for_schema(arguments[0], base_schema=base_schema)
-            return marshmallow.fields.List(child_type, **metadata)
+            child_type = field_for_schema(
+                arguments[0], base_schema=base_schema
+            )
+            list_type = generic_type_mapping[List]
+            return list_type(child_type, **metadata)
```
lovasoa, line 415: Do we really need `DEFAULT_GENERIC_TYPE_MAPPING` ? Could we just use `type_mapping.get(List, marshmallow.fields.List)` ?

YBadiss: I don't mind using `type_mapping.get(List, marshmallow.fields.List)`, I just thought it was simpler to understand having a single location with the defaults rather than having to dive in the code to figure out what defaults are used.

But it's up to you, say the word and I remove `DEFAULT_GENERIC_TYPE_MAPPING` =]

lovasoa: I understand it's a tradeoff, but I think the code will be both easier to read and more performant with `.get(key, default)`

YBadiss: ok, updated 👍

marshmallow_dataclass/__init__.py
```diff
@@ -383,16 +384,22 @@ def field_for_schema(
     origin = typing_inspect.get_origin(typ)
     if origin:
         arguments = typing_inspect.get_args(typ, True)
+        # Override base_schema.TYPE_MAPPING to change the class used for generic types below
+        generic_type_mapping = base_schema.TYPE_MAPPING if base_schema else {}
+
```
lovasoa, line 389:
```suggestion
        # Override base_schema.TYPE_MAPPING to change the class used for generic types below
        type_mapping = base_schema.TYPE_MAPPING if base_schema else {}

```

marshmallow_dataclass/__init__.py
```diff
@@ -383,16 +384,22 @@ def field_for_schema(
...
         if origin in (list, List):
             child_type = field_for_schema(arguments[0], base_schema=base_schema)
-            return marshmallow.fields.List(child_type, **metadata)
+            list_type = generic_type_mapping.get(List, marshmallow.fields.List)
```
lovasoa, line 392:
```suggestion
            list_type = type_mapping.get(List, marshmallow.fields.List)
```

marshmallow_dataclass/__init__.py
```diff
@@ -383,16 +384,22 @@ def field_for_schema(
...
+            return list_type(child_type, **metadata)
         if origin in (tuple, Tuple):
             children = tuple(
                 field_for_schema(arg, base_schema=base_schema) for arg in arguments
             )
-            return marshmallow.fields.Tuple(children, **metadata)
+            tuple_type = generic_type_mapping.get(Tuple, marshmallow.fields.Tuple)
```
lovasoa, line 398:
```suggestion
            tuple_type = type_mapping.get(Tuple, marshmallow.fields.Tuple)
```

marshmallow_dataclass/__init__.py
```diff
@@ -383,16 +384,22 @@ def field_for_schema(
...
+            return tuple_type(children, **metadata)
         elif origin in (dict, Dict):
-            return marshmallow.fields.Dict(
+            dict_type = generic_type_mapping.get(Dict, marshmallow.fields.Dict)
```
lovasoa, line 401:
```suggestion
            dict_type = type_mapping.get(Dict, marshmallow.fields.Dict)
```

lovasoa:
Great ! Let's just rename `generic_type_mapping` to `type_mapping` (since it does not contain only generics), and we'll merge.

YBadiss:
Should be good to do then =]

lovasoa:
Thank you for your contribution, @YBadiss and @CedricCabessa  !

Issue #67, opened 2020-01-29
Union dump not working


elvinlemmens:
Hi,
I have noticed the dumping of a Union type is returning an empty object:

```py
from marshmallow_dataclass import dataclass
from typing import List, Union

@dataclass
class Vector:
    x: float
    y: float

@dataclass
class Point:
    x: float
    z: float


@dataclass
class Geometries:
    elements: List[Union[Point, Vector]]

pts = Geometries.Schema().load({"elements": [
    {"x": 1, "y": 1},
    {"x": 1, "z": 1},
]})
schema=Geometries.Schema()
print(schema.dump(schema.load({"elements": [
    {"x": 1, "y": 1},
    {"x": 1, "z": 1},
]})))
```
gives
```py
{"elements": [{}, {}]}
```
but should again return
```py
{"elements": [
    {"x": 1, "y": 1},
    {"x": 1, "z": 1},
]}
```

YBadiss:
As far as I can tell, the issue is not the Union but the List:

```python3
from marshmallow_dataclass import dataclass
from typing import Union

@dataclass
class Vector:
    x: float
    y: float

@dataclass
class Point:
    x: float
    z: float


@dataclass
class Geometries:
    single: Union[Point, Vector]
    multi: List[Union[Point, Vector]]

schema=Geometries.Schema()
pts = schema.load({"multi": [
    {"x": 1, "y": 1},
    {"x": 1, "z": 1}
], "single": {"x": 1, "z": 1}})
print(schema.dump(pts))  # {'single': {'z': 1.0, 'x': 1.0}, 'multi': [{}, {}]}
```

elvinlemmens:
Indeed, thanks for noticing. A normal List[Point] works fine for example. It's the combo of Union and List that is causing th eissue.

YBadiss:
Actually it seems like there are multiple issues, the one above with `List[Union]`, but also with simple `Union`. For example doing the following will fail on the second assert

```python3
@dataclass
class Geometries:
    single: Union[Point, Vector]

schema = Geometries.Schema()

res = schema.dump(Geometries(single=Point(x=1, z=2)))
# OK
assert res == {"single": {"x": 1.0, "z": 2.0}}, res

res = schema.dump(Geometries(single=Vector(x=1, y=2)))
# NOK, only "x" is dumped but that's just because both Vector and Point have an "x" attribute
assert res == {"single": {"x": 1.0, "y": 2.0}}, res
```

[...]

PR #86, opened 2020-05-13
use polyfield for union

CedricCabessa:
marshmallow-union is not supported anymore and have some known issues
see:
https://github.com/lovasoa/marshmallow_dataclass/issues/67

Author advise to switch to marshmallow-polyfield

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ pip3 install marshmallow-dataclass
 You may optionally install the following extras:
 
 - `enum`, for translating python enums to [marshmallow-enum](https://github.com/justanr/marshmallow_enum).
-- `union`, for translating python [`Union` types](https://docs.python.org/3/library/typing.html#typing.Union) into [`marshmallow-polyfield`](https://pypi.org/project/marshmallow-polyfield/) fields.
+- `union`, for translating python [`Union` types](https://docs.python.org/3/library/typing.html#typing.Union) into [`marshmallow-union`](https://pypi.org/project/marshmallow-union/) fields.
+- `union-polyfield`, is an alternative option for Union types using [`marshmallow-polyfield`](https://pypi.org/project/marshmallow-polyfield/)
 
 ```shell
 pip3 install "marshmallow-dataclass[enum,union]"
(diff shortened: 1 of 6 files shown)


MaxRandom:
Hi @lovasoa!
Could you review this PR, please? Really need it :)

YBadiss:
Would be very useful to us! 👍

CedricCabessa:
hey @lovasoa :wave: 

Have you got a chance to look at this patch?

If you're afraid that it will break existing code, I was thinking that maybe we can keep marshmallow_union and use marshmallow_polyfield only if the module is available. This way it will be transparent to user and we can get more feedback until we are really confident.

Let me know what you think and how we can help :-)

No pressure here, cheers

lovasoa:
Hello, and sorry for not answering earlier. I had a look at the pull request. I'm not sure integrating with marshmallow-polyfield is a great idea. It is a ~90 LOC library that doesn't handle Fields, only Schemas. Creating an intermediary schema with a single field works, but it requires more code than just writing the whole feature from scratch, and I don't think this is the kind of code we want to be maintaining long term. I think the path forward is to fix existing bugs in [python-marshmallow-union](https://github.com/adamboche/python-marshmallow-union), and add support for type-based disambiguation to it. If it's not possible, than we should create a separate marshmallow union library that supports disambiguating fields based on their type. This library would be generally useful even outside of marshmallow-dataclass. 

We could then write code such as : 

```python3
UnionField([
  TypeCheckingField(int, Integer)
  TypeCheckingField(str, String)
])
```

CedricCabessa:
Yes `marshmallow-polyfield` is not ideal and I agree a clean marmallow library could be better.
However this will take time, and I far as I understand, `marshmallow-union` is [dead](https://github.com/adamboche/python-marshmallow-union#warning)

In the meantime, I can propose another solution (see last commit)
 * If marsmallow-polyfield is present, just use it. This help in some case as the unit test show.
 * if marsmallow-polyfield is not present, we stick to marshmallow-union, so no user should break and nobody have to change dependencies.

lovasoa:
I opened an issue in marshmallow-union to offer help with the maintenance: https://github.com/adamboche/python-marshmallow-union/issues/34

As I was saying, I really don't think writing a wrapper for polyfield is worth it at all: writing the serialization and deserialization logic from scratch would represent less code and be less hacky.

CedricCabessa:
ok, thank you
I'll follow the issue closely :-)
Feel free to close the PR if you want

lovasoa:
I think the tests you wrote are good. I'll merge them when we implement a solution to this issue.

PR #93, opened 2020-07-26
Improve how union fields are handled

### LINKED ISSUE #67: Union dump not working

lovasoa:
Closes #86
Closes #67

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -34,7 +34,6 @@ Full example::
       })
       Schema: ClassVar[Type[Schema]] = Schema # For the type checker
 """
-import dataclasses
 import inspect
 from enum import EnumMeta
 from functools import lru_cache
@@ -54,6 +53,7 @@ from typing import (
     overload,
 )
 
+import dataclasses
 import marshmallow
 import typing_inspect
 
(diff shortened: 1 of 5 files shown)


lovasoa:
This is an alternative to #86. 

 - **pros** : It avoids the dependency to polyfield, the hack of having to define a one-field schema for each union field, and most importantly the need to run *both* serialization and deserialization every time a field is serialized or deserialized.
 - **cons** : `Union[int,str]` will deserialize `"42"` as `42`. This is not technically incorrect, but may be surprising. This is because [`marshmallow.fields.Integer`](https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.Integer) accepts string representations of integers. More generally, the types are checked only after deserialization, so if a field in the union accepts a certain input and generates a valid output for it, then it will be used.    

@CedricCabessa, let me know what you think.

CedricCabessa:
Hey @lovasoa , thank you for this it look awesome!

Just about 

> This is easy to work around by creating strict versions of fields. For instance, fields.Integer accepts a strict parameter that can be set to True.

My solution is the following:

```
class StrictInt(Integer):
￼    def __init__(self, *args, **kwargs):
￼        kwargs["strict"] = True
￼        super().__init__(*args, **kwargs)

class CustomSchemaStrict(marshmallow.Schema):
    TYPE_MAPPING = {
        **marshmallow.Schema.TYPE_MAPPING,
        int: StrictInt,
    }

@dataclass(base_schema=CustomSchemaStrict)
class MyClass:
   foo: Union[int, str]
```

I'm wondering if you had something simpler in mind ?

I've done some test with your version, so far it is good :+1:

lovasoa:
It depends on your usage, but if you want to keep the general behavior of accepting strings for integers in general, and just override it in some places where it could cause confusions (to follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle)), you can define a [`NewType`](https://github.com/lovasoa/marshmallow_dataclass#custom-newtype-declarations) : 

```python
StrictInt = NewType("StrictInt", int, strict=True)

class MyClass:
   foo: Union[StrictInt, str] # This will deserialize integers to int and strings to str
   bar: int # This will still accept strings
```

CedricCabessa:
ok, understood!
LGTM

PR #98, opened 2020-09-08
Remove return function annotation from dataclass().

dan-starkware:
Match marshmallow_dataclass/init.py dataclass() with dataclasses.py dataclass().
Set frozen=True as a use-case. Running with return function annotation results in: Cannot instantiate type "Type[]" error.

---
This change is [Reviewable](https://reviewable.io/reviews/lovasoa/marshmallow_dataclass/98)

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -81,7 +81,7 @@ def dataclass(
     unsafe_hash: bool = False,
     frozen: bool = False,
     base_schema: Optional[Type[marshmallow.Schema]] = None,
-) -> Union[Type[_U], Callable[[Type[_U]], Type[_U]]]:
+):
     """
     This decorator does the same as dataclasses.dataclass, but also applies :func:`add_schema`.
     It adds a `.Schema` attribute to the class object
@@ -164,7 +164,6 @@ def add_schema(_cls=None, base_schema=None):
 def class_schema(
     clazz: type, base_schema: Optional[Type[marshmallow.Schema]] = None
 ) -> Type[marshmallow.Schema]:
-
     """
     Convert a class to a marshmallow schema
 
--- a/tests/test_mypy.yml
+++ b/tests/test_mypy.yml
@@ -37,3 +37,14 @@
         name: str
 
     user = User(id=4, name='Johny')
+- case: marshmallow_dataclass_type_inference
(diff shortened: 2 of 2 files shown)


lovasoa:
I'm not sure I understand the motivation for this change. You can always set `type: ignore` for the cases that are not currently manageable by mypy, but removing the type annotation altogether looks like the open door to more undetected type errors.

dan-starkware (commented):
As you can see in the test, when I want to pass arguments to dataclass (like frozen), mypy throws an error. If I type:ignore in this case, I won't get the safety of regular dataclass mypy verifications (like field names and types).
This change conforms to how python3's dataclass is written - they do no write a type annotation in the decorator.

Mypy is able to detect the type of `dataclass(frozen=True)`, but not `marshmallow_dataclass.dataclass(frozen=True)` (before this change), because the type annotation confuses it, and I can try to explain why I think this happens and why adding the annotation is actually less informative for mypy.

When you supply the type annotation, mypy accepts it as the type, and in this case, a union. Without type annotation, it is able to follow the logic inside the function, and understand exactly the right type (without the union).

> *[Reviewable](https://reviewable.io/reviews/lovasoa/marshmallow_dataclass/98#-:-MGrPbZ15oZ9YRpf90Bj:bsua7bf)* status: 0 of 2 files reviewed, all discussions resolved

lovasoa:
OK, then ! Can you complete your mypy test to check that mypy still correctly types the dataclass and its fields ? (Using `reveal_type`)

dan-starkware (commented):
Hmm.. It seems I was wrong, and mypy cannot infer the type of dataclass(frozen=True)  (It is Any).
However, I think it may still be preferable, since 1. official dataclasses implementation has no type annotation there. 2. Without it, it requires using type: ignore everywhere I pass parameters to marshmallow_dataclass.dataclass(), like frozen=True.

WDYT?

> *[Reviewable](https://reviewable.io/reviews/lovasoa/marshmallow_dataclass/98#-:-MGsHMB0ETGb0JaKLZ8L:b9rzy28)* status: 0 of 2 files reviewed, all discussions resolved

lovasoa:
I think it's way less dangerous to require programmers to explicitly silent wrongful errors than to completely remove type inference and potentially let type errors go through.

Issue #79, opened 2020-04-27
Unexpected interaction between Union and Optional


ekhahniii:
Am trying to compose Union and Optional but am seeing several unexpected behaviors. Hopefully they are captured by the below examples. Can you please let me know what I'm missing?

Common setup for observations below:
```python
from dataclasses import field
from typing import Optional, Union

from marshmallow_dataclass import dataclass
from marshmallow.validate import Equal


@dataclass
class C:
    _type: str = field(metadata=dict(validate=Equal('C')), default='C')


@dataclass
class B:
    _type: str = field(metadata=dict(validate=Equal('B')), default='B')


@dataclass
class A:
    x: Optional[Union[B, C]]
```

This fails: 
```python
raw = {'x': {'_type': 'C'}}
a = A.Schema().load(raw)
```
`marshmallow.exceptions.ValidationError: {'x': {'_type': ['Must be equal to B.']}}`.

This succeeds:
```python
raw = {'x': {'_type': 'B'}}
a = A.Schema().load(raw)
```

The following produce identical results for the above usages:
- `x: Optional[Union[B, C]] = None`
- `x: Union[B, C, None]`
- `x: Union[B, C, None] = None`

PR #39, opened 2019-09-20
Add mypy and update typings

sloria:
Since this package distributes types, it'd be good to ensure that they're correct. This adds mypy to the pre-commit config and updates type annotations to pass mypy's type checks.

PR #100, opened 2020-09-21
[ISSUE-79] Unexpected interaction between Union and Optional.

iAndriy:
Address Issue #79

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -442,14 +442,16 @@ def field_for_schema(
                 values=field_for_schema(arguments[1], base_schema=base_schema),
                 **metadata,
             )
-        elif typing_inspect.is_optional_type(typ):
-            subtyp = next(t for t in arguments if t is not NoneType)  # type: ignore
-            # Treat optional types as types with a None default
-            metadata["default"] = metadata.get("default", None)
-            metadata["missing"] = metadata.get("missing", None)
-            metadata["required"] = False
-            return field_for_schema(subtyp, metadata=metadata, base_schema=base_schema)
         elif typing_inspect.is_union_type(typ):
+            if typing_inspect.is_optional_type(typ):
+                metadata["default"] = metadata.get("default", None)
+                metadata["missing"] = metadata.get("missing", None)
+                metadata["required"] = False
+            subtypes = [t for t in arguments if t is not NoneType]  # type: ignore
+            if len(subtypes) == 1:
+                return field_for_schema(
+                    subtypes[0], metadata=metadata, base_schema=base_schema
+                )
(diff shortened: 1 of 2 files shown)


iAndriy:
@lovasoa Hey Ophir, that PR address use case when group of types supplied to `Union` has None type within it's values:
```
Union[TypeA, TypeB, None]
```

marshmallow_dataclass/__init__.py
```diff
@@ -442,14 +442,16 @@ def field_for_schema(
...
-            metadata["default"] = metadata.get("default", None)
-            metadata["missing"] = metadata.get("missing", None)
-            metadata["required"] = False
-            return field_for_schema(subtyp, metadata=metadata, base_schema=base_schema)
         elif typing_inspect.is_union_type(typ):
+            if typing_inspect.is_optional_type(typ):
+                metadata["default"] = metadata.get("default", None)
+                metadata["missing"] = metadata.get("missing", None)
+                metadata["required"] = False
+            subtypes = [t for t in arguments if t is not NoneType]  # type: ignore
```
lovasoa, line 450: Why do we need a `# type: ignore` here ?

lovasoa:
Could you also add the example you mentioned in #79 to the tests ? Trying to deserialize `{'x': {'_type': 'C'}}` with your `A` Schema.

marshmallow_dataclass/__init__.py
iAndriy: thanks, I've removed it.

iAndriy: @lovasoa Actually it's needed, otherwise mypy hook fails.

iAndriy:
@lovasoa I've added new test case which covers case Optional[Union[A, B]] case.

marshmallow_dataclass/__init__.py
lovasoa: So, why is it needed ? What is the type error ?

iAndriy: it's got here as part of modification of ```subtyp = next(t for t in arguments if t is not NoneType)  # type: ignore``` which were in place before. Do you want to dig into that ?

iAndriy: Found pr that line originally came in: https://github.com/lovasoa/marshmallow_dataclass/pull/39/files#diff-be67bc8d0a9f634f273c1261669f88caR392

lovasoa (approved):

CedricCabessa:
hey, thanks for the fix. It is useful for us!
However, serialization is broken in some case, here is another patch https://github.com/lovasoa/marshmallow_dataclass/pull/107

Issue #103, opened 2020-09-24
mypy error with marshmallow_dataclass.dataclass decorator with args: Type[<nothing>]


wpromatt:
mypy complains when using the helper dataclass decorator with arguments.
```
from typing import ClassVar
from typing import Type

from marshmallow import Schema
from marshmallow_dataclass import dataclass


@dataclass(order=True)
class Point:
    x: float
    y: float
    Schema: ClassVar[Type[Schema]] = Schema
```

**mypy error:**
```
scratch.py: note: In class "Point":
scratch.py:8:2: error: Argument 1 has incompatible type "Type[Point]"; expected "Type[<nothing>]"  [arg-type]
    @dataclass(order=True)
     ^
scratch.py:9:1: error: Cannot instantiate type "Type[<nothing>]"  [misc]
    class Point:
    ^
Found 2 errors in 1 file (checked 1 source file)
```

_Overload_:

```
@overload
def dataclass(
    *,
    repr: bool = True,
    eq: bool = True,
    order: bool = False,
    unsafe_hash: bool = False,
    frozen: bool = False,
    base_schema: Optional[Type[marshmallow.Schema]] = None,
) -> Callable[[Type[_U]], Type[_U]]:
    ...
```

#104

PR #104, opened 2020-09-24
make mypy happy when using marshmallow_dataclass.dataclass decorator …

wpromatt:
#103 add overloads for `dataclass` decorator


lovasoa:
Thank you for your contribution !

PR #106, opened 2020-09-25
add conda recipe

ElenaBossolini:
Adding conda recipe for marshmallow_dataclass with both extras (enum, union),  but excluding extra packages for docs, lints and tests. 

I have set a `conda_name` and a `pypi_name` because naming convention of conda packages is with hyphens rather than with underscores.  

I have created the package and run the installation on an ad-hoc environment from my local channel and it worked fine. But I think it would be good if you can repeat the procedure. 
1) to build the package:
run `conda build conda_recipe -c defaults -c conda-forge`  from the `marshmallow_dataclass` parent repository
2) refresh the index in your local channel by typing `conda index`
3) create and ah-hoc env such as `conda create -n marshmallow_dc_test`
4) install the package by running `conda install -n marshmallow_dc_test marshmallow**-**dataclass -c <your_local_channel> -c defaults -c conda-forge` 
This should do the game. You can check by typing  `conda list` whether the extra packages have been included as well.

### DIFF
--- a/conda_recipe/meta.yaml
+++ b/conda_recipe/meta.yaml
@@ -1,0 +1,50 @@
+{% set pypi_name = "marshmallow_dataclass" %}
+{% set conda_name = "marshmallow-dataclass" %}
+{% set version = "8.0.0" %}
+
+package:
+  name: "{{ conda_name }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ pypi_name[0] }}/{{ pypi_name }}/{{ pypi_name }}-{{ version }}.tar.gz"
+  sha256: "cde644b272c2284f1212f0e0d25bdb9eaf7d58f1423ae82b712ba43db229d618"
+
+build:
+  number: 0
+  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv "
+
+requirements:
+  host:
+    - dataclasses
+    - marshmallow >=3.0.0,<4.0
+    - marshmallow-enum
+    - pip
+    - python
+    - typeguard
+    - typing_inspect
+  run:
+    - dataclasses
+    - marshmallow >=3.0.0,<4.0
+    - marshmallow-enum
+    - python >=3.6
+    - typeguard
+    - typing_inspect
+
+test:
+  imports:
+    - marshmallow_dataclass
+
+about:
+  home: "https://github.com/lovasoa/marshmallow_dataclass"
+  license: MIT Expat License
+  license_family: MIT
+  license_file: LICENSE
+  summary: "Python library to convert dataclasses into marshmallow schemas."
+  doc_url: "https://lovasoa.github.io/marshmallow_dataclass/html/index.html"
+  dev_url: "https://github.com/lovasoa/marshmallow_dataclass"
+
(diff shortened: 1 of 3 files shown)


conda_recipe/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+{% set conda_name = "marshmallow-dataclass" %}
+{% set version = "8.0.0" %}
+
+package:
+  name: "{{ conda_name }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ pypi_name[0] }}/{{ pypi_name }}/{{ pypi_name }}-{{ version }}.tar.gz"
+  sha256: "cde644b272c2284f1212f0e0d25bdb9eaf7d58f1423ae82b712ba43db229d618"
```
lovasoa, line 11: Where does this hash come from ? Does it need to be manually updated on every release ?

conda_recipe/meta.yaml
```diff
@@ -0,0 +1,50 @@
+{% set pypi_name = "marshmallow_dataclass" %}
+{% set conda_name = "marshmallow-dataclass" %}
+{% set version = "8.0.0" %}
```
lovasoa, line 3: Does this need to be manually updated on every release ?

lovasoa:
Maybe if this requires work to maintain, this should be automated in our [.travis.yml](https://github.com/lovasoa/marshmallow_dataclass/blob/master/.travis.yml) ?

lovasoa:
Thank you for the PR, @ElenaBossolini  !

ElenaBossolini:
Hi lovasoa, 
I have just looked at bit on the internet and it seems like that the approach we are having here is not the correct one to get the package available in conda-forge. Basically, what instead we need to do is to fork from the conda-forge repository, and there add the recipe for the `marshmallow_dataclass` package, so that conda can automatically create the artifact. 
It turns out, [see for instance this tutorial](https://conda-forge.org/#add_recipe), that the hash and version are updated automatically by a bot in conda-forge once the package recipe is in there. And if you are added as a package maintainer (which I assume you want to be?) then you will get a notification everytime a new version of the `marshmallow_dataclass` is posted in Pypi, so that you can approve the new artifact in conda-forge. 
As it seems then, this PR needs to be abandoned, and instead I will open a similar PR in conda-forge.

conda_recipe/meta.yaml
ElenaBossolini: see my comment below

conda_recipe/meta.yaml
ElenaBossolini: see my comment below

lovasoa:
Ok good!

conda_recipe/meta.yaml
ElenaBossolini: Maybe [this](https://github.com/conda-forge/staged-recipes) help (see point 2, how do I populate the hash field?)

ElenaBossolini:
Here is the link to the pull request in conda-forge [#12730](https://github.com/conda-forge/staged-recipes/pull/12730)

lovasoa:
Ok, I'm closing this, then.

PR #107, opened 2020-10-09
fix serialization for Optional type

CedricCabessa:
After https://github.com/lovasoa/marshmallow_dataclass/pull/100, the
serialization works correctly, however the deserialization can fail.

eg
```
class Dclass:
  value: Optional[Union[int, str]]
```

If we load `{"value": None}` we correctly have `Dclass(value=None)`, but
serializing back this value will give an error

```
TypeError: Unable to serialize value with any of the fields in the union:
[TypeError('type of value must be int; got NoneType instead'), TypeError('type
of value must be str; got NoneType instead')]
```

The idea of this patch is to accept None when required=False

### DIFF
--- a/marshmallow_dataclass/union_field.py
+++ b/marshmallow_dataclass/union_field.py
@@ -36,6 +36,8 @@ class Union(fields.Field):
 
     def _serialize(self, value: Any, attr: str, obj, **kwargs) -> Any:
         errors = []
+        if value is None and not self.required:
+            return value
         for typ, field in self.union_fields:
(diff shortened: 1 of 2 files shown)


lovasoa:
Thanks for the fix!

CedricCabessa:
thank @lovasoa 
if you can do a release with the recent changes, that would be awesome :pray:

lovasoa:
done https://pypi.org/project/marshmallow-dataclass/#history

PR #110, opened 2020-10-26
Add support for Literal type

### LINKED ISSUE #109: Literal field and tagged union
The following simple example, which uses a [`Literal`](https://www.python.org/dev/peps/pep-0586/), doesn't work (because `marshmallow-dataclass` doesn't seem to support `Literal`):

```python

from typing_extensions import Literal

@dataclass
class A:
    choices: Literal['x', 'y']

Schema = class_schema(A)  # ERROR
```

This is the error:

```
.../site-packages/marshmallow_dataclass/__init__.py:316: UserWarning: marshmallow_dataclass was called on the class typing_extensions.Literal['x', 'y'], which is not a dataclass. It is going to try and convert the class into a dataclass, which may have undesirable side effects. To avoid this message, make sure all your classes and all the classes of their fields are either explicitly supported by marshmallow_datcalass, or are already dataclasses. For more information, see https://github.com/lovasoa/marshmallow_dataclass/issues/51
  f"marshmallow_dataclass was called on the class {clazz}, which is not a dataclass. "
  File ".../site-packages/dataclasses.py", line 970, in fields
    fields = getattr(class_or_instance, _FIELDS)
AttributeError: '_Literal' object has no attribute '__dataclass_fields__'

During handling of the above exception, another exception occurred:

  File ".../site-packages/marshmallow_dataclass/__init__.py", line 312, in _internal_class_schema
    fields: Tuple[dataclasses.Field, ...] = dataclasses.fields(clazz)
  File ".../site-packages/dataclasses.py", line 972, in fields
    raise TypeError('must be called with a dataclass type or instance')
TypeError: must be called with a dataclass type or instance

During handling of the above exception, another exception occurred:

  File ".../site-packages/marshmallow_dataclass/__init__.py", line 323, in _internal_class_schema
    created_dataclass: type = dataclasses.dataclass(clazz)
  File ".../site-packages/dataclasses.py", line 958, in dataclass
    return wrap(_cls)
  File ".../site-packages/dataclasses.py", line 950, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
  File ".../site-packages/dataclasses.py", line 764, in _process_class
    unsafe_hash, frozen))
AttributeError: '_Literal' object has no attribute '__dataclass_params__'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 303, in class_schema
    return _internal_class_schema(clazz, base_schema)
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 344, in _internal_class_schema
    for field in fields
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 345, in <genexpr>
    if field.init
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 526, in field_for_schema
    nested_schema or forward_reference or _internal_class_schema(typ, base_schema)
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 327, in _internal_class_schema
    f"{getattr(clazz, '__name__', repr(clazz))} is not a dataclass and cannot be turned into one."
TypeError: typing_extensions.Literal['x', 'y'] is not a dataclass and cannot be turned into one.
```

I think `Literal` is quite useful, e.g. for creating tagged unions. Is there any chance to add support for `Literal` to `marshmallow-dataclass`?

sisp:
This PR adds support for the [`Literal`](https://www.python.org/dev/peps/pep-0586/) type.

This is my first contribution to `marshmallow-dataclass`. Let me know if you agree with the implementation and if there's anything missing.

Fixes #109.

### DIFF
--- a/tests/test_class_schema.py
+++ b/tests/test_class_schema.py
@@ -146,9 +146,8 @@ class TestClassSchema(unittest.TestCase):
         self.assertEqual(A(data="a"), schema.load({"data": "a"}))
         self.assertEqual(schema.dump(A(data="a")), {"data": "a"})
         for data in ["b", 2, 2.34, False]:
-            self.assertRaises(
-                ValidationError, lambda data=data: schema.load({"data": data})
-            )
+            with self.assertRaises(ValidationError):
+                schema.load({"data": data})
 
     def test_literal_multiple_types(self):
         @dataclasses.dataclass
@@ -160,9 +159,8 @@ class TestClassSchema(unittest.TestCase):
             self.assertEqual(A(data=data), schema.load({"data": data}))
             self.assertEqual(schema.dump(A(data=data)), {"data": data})
         for data in ["b", 2, 2.34, False]:
-            self.assertRaises(
-                ValidationError, lambda data=data: schema.load({"data": data})
-            )
+            with self.assertRaises(ValidationError):
+                schema.load({"data": data})
 
     def test_validator_stacking(self):
         # See: https://github.com/lovasoa/marshmallow_dataclass/issues/91
--- a/tests/test_class_schema.py
+++ b/tests/test_class_schema.py
@@ -3,6 +3,11 @@ import unittest
 from typing import Any
 from uuid import UUID
 
+try:
+    from typing import Literal  # type: ignore
+except ImportError:
+    from typing_extensions import Literal
+
 import dataclasses
 from marshmallow import Schema, ValidationError
 from marshmallow.fields import Field, UUID as UUIDField, List as ListField, Integer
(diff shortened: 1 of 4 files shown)


tests/test_class_schema.py
```diff
@@ -132,6 +137,33 @@ class A:
...
+        class A:
+            data: Literal["a"]
+
+        schema = class_schema(A)()
+        self.assertEqual(A(data="a"), schema.load({"data": "a"}))
+        self.assertEqual(schema.dump(A(data="a")), {"data": "a"})
+        for data in ["b", 2, 2.34, False]:
+            self.assertRaises(
+                ValidationError, lambda data=data: schema.load({"data": data})
+            )
```
lovasoa, line 151: This is not very readable. Maybe use assertRaises as a context manager ?

```python
with self.assertRaises(ValidationError):
    schema.load({"data": data})
```

?

lovasoa (changes requested):
This looks very good. Just fix the type error on CI, and maybe use `assertRaises` as a context manager (see my comment below), and I'll merge.

https://travis-ci.org/github/lovasoa/marshmallow_dataclass/jobs/739035751

tests/test_class_schema.py
sisp: Good idea, I've made the change as you suggested.

sisp:
Thanks for your feedback! I've made the changes accordingly.

I hope the type error is fixed now. I've decided to now ignore the type-checking of the fallback import. The underlying problem seems to have no perfect solution though (e.g. https://stackoverflow.com/questions/59037244/mypy-incompatible-import-error-for-conditional-imports). If the tests pass now, are you happy with this one?

sisp:
@lovasoa The tests are passing now. Is there anything you would like me to do differently?

lovasoa:
Sorry for the delay. Thank you for your contribution !

lovasoa:
Your code is now on pypi as v8.2.0 https://pypi.org/project/marshmallow-dataclass/

sisp:
Awesome, thanks! 🙂

PR #112, opened 2020-10-29
To dict, enum by value, base schema

dliangsta:


lovasoa:
@dliangsta : What does this PR do ?

Issue #114, opened 2020-11-20
Make typing easier for custom types defined by marshmallow (Email, Url, etc.)


liberforce:
Currently, users that use type checking need to redefine types using `NewType`. Maybe `marshmallow_dataclass` could define these for the types supported by marshmallow?

I imagine something along the lines of

```
from dataclasses import field 
from marshmallow_dataclass import dataclass
from marshmallow_dataclass.typing import Email

@dataclass
class ContactInfo:
    mail: Email = field(default="anonymous@example.org")
```

lovasoa:
Yes, that's a good idea. Do you want to make a pull request?

liberforce:
I've been giving it a look, but I'm a bit puzzled about which types are required and which are not.

For example, `Email` and `Url`, IP, IPv4, IPv6 seem obviously needed.

However, from my understanding, `UUID` or `DateTime` are not required, as the python library types (uuid.UUID and datetime.datetime) can be used to perform the checks. Do you see a use case where other types like Integer, String, and others could be needed?

lovasoa:
Yes, it would be good to have an Email and Url type.

 I think marshmallow currently doesn't support the native ipaddress module, but it would be great if it could. Better than having our own type.

UUID, datetime, int, str, and others are already handled by standard library types.

liberforce:
Right, all the IP stuff is missing from `TYPES_MAPPING` in marshmallow. The way it's done, they could map `IPv4` and `IPv6`, but I'm unsure how that could be done for `IP`.

I opened an issue on marshmallow for that: https://github.com/marshmallow-code/marshmallow/issues/1695

PR #115, opened 2020-11-24
Declare common types for type checking

liberforce:
Prevents each consumer declaring them on their side.

See #114


lovasoa:
you should probably update https://github.com/lovasoa/marshmallow_dataclass/blob/master/tests/test_mypy.yml rather than creating a new .py test file on which our mypy plugin is not going to be applied. the README should also be updated.

lovasoa:
Thank you for your contribution, @liberforce !

Issue #119, opened 2020-12-23
Marshmallow 3.10.0 warning on using explicit `metadata=...` 


bonastreyair:
When using marshmallow version 3.10.0 this warning appears:

```
RemovedInMarshmallow4Warning: Passing field metadata as a keyword arg is deprecated. 
Use the explicit `metadata=...` argument instead.
```

Here is the commit where they have changed the behavior. https://github.com/marshmallow-code/marshmallow/commit/013abfd669f64446cc7954d0320cf5f1d668bd49

I understand this is something that needs to be addressed on the `marshmallow_dataclass` side, right?

lovasoa:
Can you give an example of code that triggers this?

bonastreyair:
Given this `test_m.py` file:

```python
from dataclasses import dataclass, field
import marshmallow_dataclass
import marshmallow

@dataclass
class Foo:
    bar: float = field(
        metadata={
            "description": "test",
        }
    )

def test_create_schema():
    schema=marshmallow_dataclass.class_schema(Foo)()
```

Running pytest on that file shows this warning:
```
$ pytest test_m.py

========================== test session starts ==========================
platform darwin -- Python 3.8.1, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
rootdir: current/path/test
collected 1 item                                                        

test_m.py .                                                       [100%]

=========================== warnings summary ============================
test_m.py::test_create_schema
  /Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/marshmallow/fields.py:195: RemovedInMarshmallow4Warning: Passing field metadata as a keyword arg is deprecated. Use the explicit `metadata=...` argument instead.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/warnings.html
===================== 1 passed, 1 warning in 0.06s ======================
```

If no metadata is set then warning does not appear.

bonastreyair:
Using using `"metadata"` as a key inside  `field(metadata={...})` seems to solve the issue...

No warnings when pytesting the following `test_m.py` file:
```python
from dataclasses import dataclass, field
import marshmallow_dataclass

@dataclass
class Foo:
    bar: float = field(
        metadata={
            "required": False,
            "default": 0.0,
            "metadata": {
                "description": "test",
            }
        }
    )

def test_create_schema():
    schema = marshmallow_dataclass.class_schema(Foo)()
```

lovasoa:
[...]

PR #123, opened 2021-01-21
Do not pass "description" argument to marshmallow.Field

noirbee:
This should also be a fix for the initial issue reported in
https://github.com/lovasoa/marshmallow_dataclass/issues/119 (`description` is the problem here, not
the other arguments)

Note also the use of PYTHONWARNINGS=error to help detect these types of errors in the future.


lovasoa (approved):

lovasoa:
Thanks !

PR #129, opened 2021-03-25
Add MIT license trove classifier

joshfriend:
Allows the license to be detected properly by tools like `pip-licenses`


lovasoa:
Thanks for the contribution

Issue #51, opened 2019-11-06
When type hint is not recognized (e.g. `pd.Timestamp`), we recursively try to convert into a dataclass which mutates the constructor throughout the codebase


evanfwelch:
Reproducible error:
```
import pandas as pd
import dataclasses

import marshmallow_dataclass

@dataclasses.dataclass
class Something:
    date_i_care_about: pd.Timestamp

print(f"here's a timestamp {pd.Timestamp('2017-01-01')}")

# this mutates the constructor for pd.Timestamp by recursively trying to convert to dataclass
marshmallow_dataclass.class_schema(Something)

print(f"here's a timestamp {pd.Timestamp('2017-01-01')}")

```

versions:
pandas 0.23.4
marshmallow 2.x
marshmallow_dataclass: 0.6.6

@lovasoa what is the intution behind trying to convert `clazz` to dataclass here after we fail to get the fields?

https://github.com/lovasoa/marshmallow_dataclass/blob/master/marshmallow_dataclass/__init__.py#L281

I feel that an error (or some kind of `strict` mode) would be more helpful

lovasoa:
marshmallow_dataclass allows you to have simple classes, like : 

```python
class A:
    a: int

class B:
    b: A

marshmallow_dataclass.class_schema(B)
```

In order to create a schema for B, it also has to create a schema for A. And in order to apply `dataclasses.fields` to A, it has to make it a dataclass: https://github.com/lovasoa/marshmallow_dataclass/blob/master/marshmallow_dataclass/__init__.py#L281

I don't think it's possible to fix this issue without breaking backward compatibility...
What behavior would you expect for your example above ?

lovasoa:
If you want a custom type in your class, the best way is to use a [NewType declaration](https://github.com/lovasoa/marshmallow_dataclass#custom-newtype-declarations)

evanfwelch:
Ok I see your point. In my (and my colleagues) use case, we would already have made both `A` and `B` a `dataclass` (because we want all the benefits of dataclasses in our application code), and then when it comes to serializing an instance of B, we'd make generate a marshmallow schema by doing:

`marshmallow_dataclass.class_schema(B).dump(...)`

However I guess you've primarily designed for the case where folks are using the `marshmallow_dataclass.dataclass` decorator to do both at once....

[...]

PR #130, opened 2021-03-30
Increase warning visibility, suggest using field metadata

prihoda:
Increases visibility of warning from https://github.com/lovasoa/marshmallow_dataclass/issues/51


lovasoa (approved):
Looks good to me !

Issue #131, opened 2021-04-03
Add support for Mapping, Sequence, Set and FrozenSet


anis-campos:
The goal here is to go a step further than #120 and add support for the other commonly used collection types. 

Why do I even bother ? 
---
the main goal here is to "reintroduce" the Immutable fields in the dataclass. Currently we are limited to `List, Dict, Tuple`. This is mostly fine, but `List` and `Dict` are mutable and give the overall "meh" result once associated with `dataclass(frozen=true)`.
Thus, among those in the list, `Sequence` and `Mapping` are my must have. `Set` and `FrozenSet` are here just for completness.

Notes:
---
I'm not very sure how to handle the new fields needed to implement this. 

The implementation is pretty straight forward as it's just a special case of `marshmallow.fields.List`, by simply changing the collection type used in the `_deserialize`. 

But, just like the `Union` field, should those be considered extras ?

anis-campos:
Seems related to #99

PR #132, opened 2021-04-03
Add support for Mapping, Sequence, Set and FrozenSet

anis-campos:
With this, most used collection types are supported in the dataclass.

fixes lovasoa/marshmallow_dataclass#131

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -1,3 +1,4 @@
+import itertools
 import unittest
 from typing import Optional, Sequence, Set, FrozenSet, Mapping
 
@@ -122,6 +123,9 @@ class TestSequenceField(unittest.TestCase):
 
 
 class TestSetField(unittest.TestCase):
+    def assertEqualAsSet(self, a, b):
+        self.assertEqual(set(a), set(b))
+
     def test_simple(self):
         @dataclass
         class IntSet:
@@ -131,7 +135,7 @@ class TestSetField(unittest.TestCase):
         data_in = {"value": list(range(5))}
         loaded = schema.load(data_in)
         self.assertEqual(loaded, IntSet(value=set(range(5))))
-        self.assertCountEqual(schema.dump(loaded)["value"], data_in["value"])
+        self.assertEqualAsSet(schema.dump(loaded)["value"], data_in["value"])
 
     def test_set_no_arg(self):
         @dataclass
@@ -144,7 +148,7 @@ class TestSetField(unittest.TestCase):
         data_in = {"value": [1, "2", 3.0, (1, 2, 3)]}
         loaded = schema.load(data_in)
         self.assertEqual(loaded, AnySet(value={1, "2", 3.0, (1, 2, 3)}))
-        self.assertCountEqual(schema.dump(loaded)["value"], data_in["value"])
+        self.assertEqualAsSet(schema.dump(loaded)["value"], data_in["value"])
 
     def test_optional_set(self):
         @dataclass
@@ -161,7 +165,7 @@ class TestSetField(unittest.TestCase):
         data_in = {"value": list(range(5))}
         loaded = schema.load(data_in)
         self.assertEqual(loaded, IntOptionalSet(value=set(range(5))))
-        self.assertCountEqual(schema.dump(loaded)["value"], data_in["value"])
+        self.assertEqualAsSet(schema.dump(loaded)["value"], data_in["value"])
 
     def test_set_of_optional(self):
         @dataclass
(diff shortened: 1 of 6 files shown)


lovasoa (changes requested):
Great, this is a useful addition ! 
Could you please add tests for your new field types ? 
You can use  schema.load and schema.dump, loading data from json to a dataclass and back.

anis-campos:
Let me know if the test are not good enough

tests/test_collection.py
```diff
@@ -0,0 +1,366 @@
...
+            "has",
+            "only",
+            "1",
+            "chance",
+            "in",
+            "40M",
+            "to",
+            "keep",
+            "it's",
+            "ordered",
```
lovasoa, line 142:
```suggestion
            "its",
            "order",
```

tests/test_collection.py
```diff
@@ -0,0 +1,366 @@
...
+        )
+        dump = str_schema.dump(StrSequence(value=set(sequence)))
+        self.assertNotEqual(
+            dump,
+            {"value": sequence},
+        )
```
lovasoa, line 148: You shouldn't assertNotEqual here. You should probably assert that `dump["value"] == set(sequence)` instead

tests/test_collection.py
```diff
@@ -0,0 +1,366 @@
...
+
+        @dataclass
+        class IntSequence:
+            value: Sequence[int]
+
+        int_schema = IntSequence.Schema()
+
+        # and to make it worse, int sets are a valid sequence that keep order....
+        dump = int_schema.dump(IntSequence(value={1, 2, 3, 4, 5}))
```
lovasoa, line 157: Is that guaranteed by the spec or just an implementation detail ?

tests/test_collection.py
```diff
@@ -0,0 +1,366 @@
...
+            value: Optional[Set[int]]
+
+        schema = IntOptionalSet.Schema()
+        for data_in in ({"value": list(range(5))}, {"value": []}, {"value": None}):
+            dumped_value = schema.dump(schema.load(data_in))["value"]
+            value = schema.dump(schema.load(data_in))["value"]
+            if value is None:
+                self.assertIsNone(dumped_value)
+            else:
+                self.assertCountEqual(dumped_value, value)
```
lovasoa, line 198: What are you trying to test here ?
Why computing `schema.dump(schema.load(data_in))["value"]` twice ?

tests/test_collection.py
```diff
@@ -0,0 +1,366 @@
...
+    def test_set_of_optional(self):
+        @dataclass
+        class OptionalIntSet:
+            value: Set[Optional[int]]
+
+        schema = OptionalIntSet.Schema()
+        for data_in in ({"value": set(range(5))}, {"value": {1, None, 2}}):
+            self.assertCountEqual(
+                schema.dump(schema.load(data_in))["value"], data_in["value"]
+            )
```
lovasoa, line 209:
```suggestion
            self.assertEqual(
                set(schema.dump(schema.load(data_in))["value"]), data_in["value"]
            )
```

lovasoa (changes requested):

tests/test_collection.py
anis-campos: Agreed. Not a very deterministic test as it is. I wanted to point out the fact that dumping set will yield a random sorted list, but I guest that checking that is out of scope.

tests/test_collection.py
anis-campos: It's more a find-out while testing. IMO it comes down to how a set is implemented were ints will stay in order, i suppose because d`hash(int)=int`.

But again, looking back, this whole test seams out of scope. I will remove it

tests/test_collection.py
anis-campos: Yep, overly complicated test that missed the point. I just wanted to test the None case here I guess. I will simplify.

lovasoa:
I made a few changes.
If you agree with everything and have nothing to add, I'll merge and release v8.4.0

anis-campos:
> I made a few changes.
> If you agree with everything and have nothing to add, I'll merge and release v8.4.0

Yes that's great. Thanks for the rewies and help !

lovasoa (approved):

lovasoa:
[PyPI seems to be down at the moment](https://status.python.org/incidents/rw171ylf8jw3). I'll push the upgrade when it comes back online.

lovasoa:
WOW ! There was a syntax error that slept through to our CI that prevented it to run for this PR. This actually broke compatibility with almost all of our supported python versions. I'm reverting this.
https://github.com/lovasoa/marshmallow_dataclass/runs/2270735343?check_suite_focus=true

anis-campos:
> WOW ! There was a syntax error that slept through to our CI that prevented it to run for this PR. This actually broke compatibility with almost all of our supported python versions. I'm reverting this.
> https://github.com/lovasoa/marshmallow_dataclass/runs/2270735343?check_suite_focus=true

😨 .  Wait, I don't see any test run in the PR. It only happens wile trying to publish master right ?

PR #133, opened 2021-04-05
Fix compatibility with older python versions

lovasoa:
#132 broke compatibility with python 3.6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # marshmallow\_dataclass change log
 
+## v8.4.1
+
+ - Fix compatibility with older python versions.
+   
+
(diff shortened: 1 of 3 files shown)


lovasoa:
Since the majority of users are on python 3.6, this is pretty critical

anis-campos:
Good to know that it wasn't too bad.  🤔  I was actually wondering why I had to fall back to `collections.abc.xxx` to check the types.

anis-campos:
TIL that there is implementation difference in the `typing_inspect.get_origin` method across the some python versions.

lovasoa:
See https://docs.python.org/3/library/stdtypes.html#types-genericalias

PR #137, opened 2021-04-25
Proposal: add union serialization strategy

Tosa95:
I would like to have the possibility to choose how unions are serialized and deserialized. The try-all-possibility strategy that is currently used is great for most of the cases, but when two or more types among the possible ones share the same schema it fails. Moreover, trying out all possible types can be time consuming.

Instead of this, my proposal is to let the user choose which strategy should be followed when serializing/deserializing a union. In particular I implemented the following two strategies:

- **Fallback**: tries all the types until it finds one that is successfully deserialized and validated (this is the default one)
- **TypeName**: during serialization adds an attribute with the name of the type that is used during deserialization to load the correct type. I let the user free to choose the key used for the type and the names used for types (if they don't specify them, class name is used by default)

The strategy to be used can be specified in metadata. If it is not specified, fallback is assumed.

The user can also implement new strategies if they want. 

There is another option to obtain what I wanted without modifying the code that is manually adding a type field to dataclasses in the union and then use validation to make deserialization fail in case the type is not the wanted one, but I found this a little hacky.

If you like this little contribution I will clean up the code and add some documentation about this new feature.

In any case thank you for this awesome little library!

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -37,6 +37,7 @@ Full example::
 import collections.abc
 import dataclasses
 import inspect
+from marshmallow_dataclass.union_field import SERIALIZATION_STRATEGY_KEY
 import warnings
 from enum import EnumMeta
 from functools import lru_cache
@@ -71,6 +72,8 @@ MEMBERS_WHITELIST: Set[str] = {"Meta"}
 
 # Max number of generated schemas that class_schema keeps of generated schemas. Removes duplicates.
 MAX_CLASS_SCHEMA_CACHE_SIZE = 1024
+
+PASSTHROUGH_METADATA_KEYS = {SERIALIZATION_STRATEGY_KEY}
 
 
 @overload
@@ -431,6 +434,9 @@ def _generic_type_add_any(typ: type) -> type:
     return typ
 
 
+def extract_passthrough_metadata(metadata):
+    return {k: v for k, v in metadata.items() if k in PASSTHROUGH_METADATA_KEYS}
+
 def _field_for_generic_type(
     typ: type, base_schema: Optional[Type[marshmallow.Schema]], **metadata: Any
 ) -> Optional[marshmallow.fields.Field]:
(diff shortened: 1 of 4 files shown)


xor-xor:
I was going to open a separate issue, but found this one so I'll write here instead.

I was scratching my head with the result of this snippet:
```python
import typing as T
from marshmallow_dataclass import dataclass as mm_dataclass

@mm_dataclass
class Foo:
    x: T.Union[float, int]

Foo.Schema().dumps(Foo(1))    

# => '{"x": 1.0}'
```
In particular, why `1` gets casted to `1.0` despite integer being a "legal" type (i.e. allowed by this union)?

This may seem like a minor thing, but actually it may have disastrous consequences - for example, there's this machine learning framework PyTorch Lightning, with params like e.g. `limit_train_batches`. Passing `1` to it means something totally different than `1.0` (integer is interpreted as the number of batches that should be taken from the training set, while float is interpreted as a fraction of the training set).

I've found that marshmallow_dataclass just tries the types in the order given by the union and continues with the first one that "just works" (the "Fallback" strategy from @Tosa95's proposal). Another strategy (preferred one in my case) could be "try all the types given by the union, and continue with the one which exactly matches the actual type of the data".

@lovasoa any thoughts on this?

lovasoa:
Unfortunately, json itself (JavaScript Object Notation), just as JavaScript, doesn't make a semantic difference between floats and integers. 1.0 and 1 mean the exact same thing in json. 

You can create two dataclasses to differentiate between the two:

```
class Float:
  name = "float"
  value: float

class Int:
  name = "int"
  value: int
```

xor-xor:
Although you're right that JSON doesn't have integers as a distinct type, it certainly supports them (see [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7159.html#section-6)), so I guess it boils down to language's implementation. For example, Python's `json` module from its standard library [distinguishes integers and floats](https://docs.python.org/3/library/json.html#encoders-and-decoders). Also:
```python
import typing as T
from marshmallow_dataclass import dataclass as mm_dataclass

@mm_dataclass
class Foo:
    x: T.Union[float, int]

foo = Foo.Schema().loads(
    Foo.Schema().dumps(Foo(1))
)
type(foo.x)
# => float

@mm_dataclass
class Foo:
    x: T.Union[int, float]

foo = Foo.Schema().loads(
    Foo.Schema().dumps(Foo(1))
)
type(foo.x)
# => int
```
Therefore, if there's only Python in my serialization-deserialization routine, then why I shouldn't be able to exploit the fact that it supports integers encoded in JSON format? Of course, I wouldn't make this as a default, but _some_ way to achieve this would be very appreciated.

xor-xor:
The problem with imposing a meaning to the order of `Union` arguments is even more apparent with `float` and `bool` types, as both of them have their distinct counterparts in JSON format (and in JavaScript as well). Consider this:
```python
import typing as T
from marshmallow_dataclass import dataclass as mm_dataclass

@mm_dataclass
class Foo:
    x: T.Union[float, bool]

foo = Foo.Schema().loads(
    Foo.Schema().dumps(Foo(0.0))
)
type(foo.x)
# => float

@mm_dataclass
class Foo:
    x: T.Union[bool, float]

foo = Foo.Schema().loads(
    Foo.Schema().dumps(Foo(0.0))
)
type(foo.x)
# => bool
```

(BTW, sorry for separate comments - it's just like some separate process in my head is analyzing this in the background and from time to time it spits out some new insights 😉)

Issue #144, opened 2021-05-26
Way to completely ignore an Optional field when dumping ?


Stepland:
Hello, first off thank you so much for making such a nice library !

I may be missing something obvious, but I'm having a hard time replicating marshmallow's original dump behavior for a field that's not required.

When using a regular marshmallow schema, if I somehow **have** to pass an object with an `a` attribute/key but don't want it to appear in the serialized output, I can use `marshmallow.missing`
```py3
>>> class Thing(Schema):
...     a = fields.String()
...
>>> Thing().dump({"a": missing})
{}
```
It's not pretty but it works ...

However the equivalent trick behaves differently with marshmallow_dataclass
```py3
>>> @dataclass
... class Thing:
...     a: Optional[str]
... 
>>> Thing.Schema().dump(Thing(a=missing))
{'a': None}
```

I don't know if this is intentional, if there's another recommended way of doing this I'd be more than happy to use that instead.

YuvalPruss:
Also asked myself the same question. Is there a way to omit those nullable values?

lovasoa:
You can use a post-dump method. See https://github.com/marshmallow-code/marshmallow/issues/633

Stepland:
Welp, completely forgot about that, sorry and thanks a ton for helping !

PR #145, opened 2021-05-26
Make Optional field `default` param default to marshmallow.missing instead of None

Stepland:
Fix proposal for #144


lovasoa:
This would break backwards compatibility. I'm not sure we want that...

PR #148, opened 2021-06-22
Fix mypy tests

sisp:
Since mypy v0.900, double-quotes instead of single-quotes are used in error messages:
- https://github.com/python/mypy/commit/96cae4b1b0c2a4298ca8cb3e86d8dced30c9fff2
- https://github.com/python/mypy/compare/v0.812...v0.900

When the last CI pipeline ran, mypy v0.812 was used which is why no errors occurred.

Also, it seems mypy v0.900 requires type stubs for the dataclasses backport needed for Python 3.6. Just wondering, should types-dataclasses be a dependency of this library or rather of the dataclasses backport library?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ CLASSIFIERS = [
 EXTRAS_REQUIRE = {
     "enum": ["marshmallow-enum"],
     "union": ["typeguard"],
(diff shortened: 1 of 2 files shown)


lovasoa:
Thanks @sisp !

PR #150, opened 2021-06-22
Add support for Final type

### LINKED ISSUE #111: Final type
The following simple example, which uses the [`Final`](https://docs.python.org/3/library/typing.html#typing.Final) type hint, doesn't work (because `marshmallow-dataclass` doesn't seem to support `Final`):

```python

from typing_extensions import Final

@dataclass
class A:
    value: Final[int]

Schema = class_schema(A)  # ERROR
```

This is the error:

```
.../site-packages/marshmallow_dataclass/__init__.py:316: UserWarning: marshmallow_dataclass was called on the class typing_extensions.Final[int], which is not a dataclass. It is going to try and convert the class into a dataclass, which may have undesirable side effects. To avoid this message, make sure all your classes and all the classes of their fields are either explicitly supported by marshmallow_datcalass, or are already dataclasses. For more information, see https://github.com/lovasoa/marshmallow_dataclass/issues/51
  f"marshmallow_dataclass was called on the class {clazz}, which is not a dataclass. "
  File ".../site-packages/dataclasses.py", line 970, in fields
    fields = getattr(class_or_instance, _FIELDS)
AttributeError: '_Final' object has no attribute '__dataclass_fields__'

During handling of the above exception, another exception occurred:

  File ".../site-packages/marshmallow_dataclass/__init__.py", line 312, in _internal_class_schema
    fields: Tuple[dataclasses.Field, ...] = dataclasses.fields(clazz)
  File ".../site-packages/dataclasses.py", line 972, in fields
    raise TypeError('must be called with a dataclass type or instance')
TypeError: must be called with a dataclass type or instance

During handling of the above exception, another exception occurred:

  File ".../site-packages/marshmallow_dataclass/__init__.py", line 323, in _internal_class_schema
    created_dataclass: type = dataclasses.dataclass(clazz)
  File ".../site-packages/dataclasses.py", line 958, in dataclass
    return wrap(_cls)
  File ".../site-packages/dataclasses.py", line 950, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
  File ".../site-packages/dataclasses.py", line 764, in _process_class
    unsafe_hash, frozen))
AttributeError: '_Final' object has no attribute '__dataclass_params__'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 303, in class_schema
    return _internal_class_schema(clazz, base_schema)
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 344, in _internal_class_schema
    for field in fields
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 345, in <genexpr>
    if field.init
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 526, in field_for_schema
    nested_schema or forward_reference or _internal_class_schema(typ, base_schema)
  File ".../site-packages/marshmallow_dataclass/__init__.py", line 327, in _internal_class_schema
    f"{getattr(clazz, '__name__', repr(clazz))} is not a dataclass and cannot be turned into one."
TypeError: typing_extensions.Final[int] is not a dataclass and cannot be turned into one.
```

I find the `Final` type quite useful for annotating a field to be read-only without runtime overhead.

sisp:
This PR adds support for the `Final` type.

Let me know if you agree with the implementation and if there's anything missing.

Closes #111.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # marshmallow\_dataclass change log
 
+- Add support for the Final type. See [#150](https://github.com/lovasoa/marshmallow_dataclass/pull/150)
+
 ## v8.4.1
 
  - Fix compatibility with older python versions.
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -579,6 +579,16 @@ def field_for_schema(
             **metadata,
         )
 
+    if typing_inspect.is_final_type(typ):
+        arguments = typing_inspect.get_args(typ)
+        if arguments:
+            subtyp = arguments[0]
+        elif default is not marshmallow.missing:
+            subtyp = type(default)
+        else:
+            subtyp = Any
+        return field_for_schema(subtyp, default, metadata, base_schema)
+
     # Generic types
     generic_field = _field_for_generic_type(typ, base_schema, **metadata)
     if generic_field:
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
     classifiers=CLASSIFIERS,
     license="MIT",
     python_requires=">=3.6",
-    install_requires=["marshmallow>=3.0.0,<4.0", "typing-inspect"],
+    install_requires=["marshmallow>=3.0.0,<4.0", "typing-inspect>=0.7.0"],
     extras_require=EXTRAS_REQUIRE,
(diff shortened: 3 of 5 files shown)


lovasoa (approved):
Thanks !

PR #151, opened 2021-06-23
Add warning about the limitations of type inference when using Final

sisp:
Unfortunately, I implemented type inference from a default value when using `Final` incorrectly in #150, i.e. it only works for simple cases, but it's actually not true type inference. I'm not sure whether it's currently possible to get the type hint inferred from the default value at runtime.

E.g.:

```python
@dataclasses.dataclass
class A:
    data: Final = dataclasses.field(default_factory=lambda: [1, "a"])  # inferred as `list[int | str]`
```

1. The implementation in #150 does not support `dataclasses.field(...)`, but this could be fixed.
2. The implementation in #150 would at best infer the default value to have the type `list` but cannot infer `List[Union[int, str]]`. In order to get the correct type, actual type inference is be needed.

This PR changes the behavior, so that a `NotImplementedError` is raised when `Final` has no argument and a default value exists. I don't think it's a big problem to not support this case because type inference is just a convenience. Instead, `Final[...]` could always be typed explicitly.

Sorry for the sloppiness in #150 ...

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -585,6 +585,11 @@ def field_for_schema(
             subtyp = arguments[0]
         elif default is not marshmallow.missing:
             subtyp = type(default)
+            warnings.warn(
+                "Support for type inference from a default value is limited and may "
+                "result in inaccurate validation. Provide a type to Final[...] to "
+                "ensure accurate validation."
+            )
         else:
             subtyp = Any
         return field_for_schema(subtyp, default, metadata, base_schema)
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -584,7 +584,9 @@ def field_for_schema(
         if arguments:
             subtyp = arguments[0]
         elif default is not marshmallow.missing:
-            subtyp = type(default)
+            raise NotImplementedError(
+                "Type inference from default value is not supported"
+            )
         else:
             subtyp = Any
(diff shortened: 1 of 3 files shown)


lovasoa:
I think the behavior of #150 is ok... Of course you cannot get a complete type definition from just a single value, but using `type` can be useful in simple cases. Maybe just add a warning in this case ?

marshmallow_dataclass/__init__.py
```diff
@@ -585,6 +585,11 @@ def field_for_schema(
             subtyp = arguments[0]
         elif default is not marshmallow.missing:
             subtyp = type(default)
+            warnings.warn(
+                "Support for type inference from a default value is limited and may "
+                "result in inaccurate validation. Provide a type to Final[...] to "
+                "ensure accurate validation."
```
lovasoa, line 591: Given this message, it will not be obvious where the issue comes from. I thing the warning should mention that it comes from marshmallow_dataclass, the name of the type, and the name of the field.

sisp: Agreed. I've pushed a new commit, but I don't know how to get the name of the field. Is there a way to get it from within `field_for_schema`?

sisp:
I've added a warning about the limitations of using a default factory according to our discussion in #152.

lovasoa (approved):
thanks and sorry for the delay !

sisp:
No worries, thanks for merging the PR! 🙂

PR #152, opened 2021-06-23
Support Final field with a default factory

sisp:
This PR adds support for using `Final` with a field and a default factory:

```python
@dataclasses.dataclass
class A:
    data: Final = dataclasses.field(default_factory=lambda: ["a", 1])
```

Extends #150.


lovasoa:
I think it would be quite unexpected for the user to see the default_factory function called in this context.
And we can't assume how expensive the function call is, and whether it has side effects or not...

Is there a context in particular where you think the advantages of this over just specifying a type in the type annotation would outweigh the cons ?

lovasoa:
yes

Issue #157, opened 2021-07-22
RemovedInMarshmallow4Warnings for `missing` and `default` arguments


mivade:
I've started seeing warnings like the following:

```
RemovedInMarshmallow4Warning: The 'missing' argument to fields is deprecated. Use 'load_default' instead.
```

and

```
The 'default' argument to fields is deprecated. Use 'dump_default' instead.
```

It looks like these are coming from `marshmallow_dataclass` itself and is from a recent change (https://github.com/marshmallow-code/marshmallow/pull/1742) affecting Marshmallow version >= 3.13.0 (released yesterday).

lovasoa:
Yes, we need to update marshmallow-dataclass for the latest marshmallow. PR welcome!

maroux:
Opened a PR for this issue.

mivade:
Thanks @maroux!

PR #158, opened 2021-07-27
Replace missing/default with {load,dump}_default

maroux:
Chose to bump minimum marshmallow version since its not a breaking change. Alternative is to add `compat.py` and check marshmallow version before creating `field` instances.

Also added `filterwarnings` pytest setting to convert all warnings into errors so we can catch this kind of problem.

Fixes lovasoa/marshmallow_dataclass#157

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # marshmallow\_dataclass change log
 
+## v8.5.0
+- Fix `default` warning coming from marshmallow. Bump minimal marshmallow version requirement to 3.13. 
 - Add support for the Final type. See [#150](https://github.com/lovasoa/marshmallow_dataclass/pull/150)
 
 ## v8.4.2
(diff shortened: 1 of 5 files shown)


lovasoa (approved):
Looks good to me.

lovasoa:
Thank you @maroux !

mivade:
Thanks! @lovasoa presumably when this lands in a release it would be a major version bump, right? Otherwise projects that are setting too strict requirements for both Marshmallow and this project will have problems.

lovasoa:
It's a minor version. What would be the problem with other projects? If they have a strict requirement on a lower marshmallow version, they just won't upgrade marshmallow-dataclass

mivade:
If I have requirements like

```
"marshmallow_dataclass>=8",
"marshmallow>=3,<4",
```

then installs would fail. Ideally only `marshmallow_dataclass` would be specified but that's not what always happens. It's a minor point though since fixing requirements is an easy enough change.

maroux:
@mivade I'm not sure why install would fail? You'd resolve to marshmallow==3.13.0 and marshmallow_dataclass==8.5.0 ?

mivade:
You're right, that was a bad example that I wrote hastily. I have however seen some overzealous cases more like `marshmallow>=3.0,<3.10` or sometimes `marshmallow==3.10.0`. Thinking about it some more though and I think that's really a bug in the consumer code (there's really no reason to explicitly require Marshmallow if you're already requiring m-dc).

PR #159, opened 2021-08-09
Allow setting `required=True` on fields of type `Optional`

dairiki:
As things currently are, `marshmallow-dataclass` [always forces `required=False`][1] on fields with `Optional` type.  It would be useful to be able to set `required=True` on such fields.

[1]: <https://github.com/lovasoa/marshmallow_dataclass/blob/05418800ccdaf158826bb34ceb12fc5c75daa266/marshmallow_dataclass/__init__.py#L495>

For example, if I write:
```py
@dataclass
class Data:
    x: Optional[int] = field(default=0, metadata={'required': True})
```
then, on deserialization, it should be required that a value for `x` — either an integer or `None` — be present.

Currently, `marshmallow-dataclass` forces `required=False` and `load_default=None` on the schema field, so that
`Data.Schema().load({})` will succeed, returning `Data(x=None)`.  I would like to be able to configure things so that, instead, `Data.Schema().load({})` raises `ValidationError("Missing data for required field")`.

This PR does that.

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -491,8 +491,9 @@ def _field_for_generic_type(
             if typing_inspect.is_optional_type(typ):
                 metadata["allow_none"] = metadata.get("allow_none", True)
                 metadata["dump_default"] = metadata.get("dump_default", None)
-                metadata["load_default"] = metadata.get("load_default", None)
-                metadata["required"] = False
+                if not metadata.get("required"):
+                    metadata["load_default"] = metadata.get("load_default", None)
+                metadata.setdefault("required", False)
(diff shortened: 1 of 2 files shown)


lovasoa (approved):
Looks good! I'll merge when all the tests finish running

lovasoa:
The tests are failing...

lovasoa (changes requested):
tests don't pass

dairiki:
Doh!  Sorry, I made a last-minute change to the type of the field in the test, and — apparently — didn't check to make sure the test still ran correctly.  Fixed.  At least here.

lovasoa (approved):

dairiki:
Thank you!

Issue #161, opened 2021-08-10
Problem with the Union type: ValueError: 'load_default' must not be set for required fields.


xor-xor:
Hi,

Recently I've came accross a problem with the `Union` type, which can be reproduced with the following snippet:

```
from typing import Union

from marshmallow_dataclass import dataclass as mm_dataclass


@mm_dataclass
class MyClass:
    field1: int = 1
    field2: Union[int, float] = 2.0
```

Evaluating it with `marshmallow-dataclass[union]==8.5.1` gives:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_21842/1436176557.py in <module>
      6 
      7 
----> 8 @mm_dataclass(frozen=True)
      9 class MyClass:
     10     field1: int = 1

~/envs/mm/lib/python3.7/site-packages/marshmallow_dataclass/__init__.py in <lambda>(cls)
    142     )
    143     if _cls is None:
--> 144         return lambda cls: add_schema(dc(cls), base_schema)
    145     return add_schema(dc, base_schema)
    146 

~/envs/mm/lib/python3.7/site-packages/marshmallow_dataclass/__init__.py in add_schema(_cls, base_schema)
    191         return clazz
    192 
--> 193     return decorator(_cls) if _cls else decorator
    194 
    195 

~/envs/mm/lib/python3.7/site-packages/marshmallow_dataclass/__init__.py in decorator(clazz)
    188     def decorator(clazz: Type[_U]) -> Type[_U]:
    189         # noinspection PyTypeHints
--> 190         clazz.Schema = class_schema(clazz, base_schema)  # type: ignore
    191         return clazz
    192 

~/envs/mm/lib/python3.7/site-packages/marshmallow_dataclass/__init__.py in class_schema(clazz, base_schema)
    316     if not dataclasses.is_dataclass(clazz):
    317         clazz = dataclasses.dataclass(clazz)
--> 318     return _internal_class_schema(clazz, base_schema)
    319 
    320 

~/envs/mm/lib/python3.7/site-packages/marshmallow_dataclass/__init__.py in _internal_class_schema(clazz, base_schema)
    360             ),
    361         )
--> 362         for field in fields
    363         if field.init
    364     )

~/envs/mm/lib/python3.7/site-packages/marshmallow_dataclass/__init__.py in <genexpr>(.0)
    361         )
    362         for field in fields
--> 363         if field.init
    364     )
    365 

[...]

PR #162, opened 2021-08-10
Proposed fix for #161

dairiki:
Here is a proposed fix for #161.

Before you merge this, you might want to sit and think on it a bit to make sure that my logic is sound, but I'm thinking that the component fields of the `Union` field should all have `required=True` set.

The `Union` field itself (in its `deserialize` and `serialize` methods) handles the logic having to do with its `required`, `load_default`, `allow_none`, and `dump_default` options.  The component union_fields are only used for serializing and deserializing non-default/non-null values.  We don't want them to fill in default values or pass through `None`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # marshmallow\_dataclass change log
 
+## v8.5.3
+
+- Fix spurious `ValueError` when defining a Union field with explicit default value
+  ([#161](https://github.com/lovasoa/marshmallow_dataclass/pull/161))
+
 ## v8.5.2
 
 - Fix spurious `TypeError` when serializing `Optional` union types with `required=True`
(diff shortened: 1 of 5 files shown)


lovasoa:
@dairiki : I think you are right, and you are making many useful PRs. Are you interested in becoming a maintainer, with push access to this repo ?

dairiki:
@lovasoa: Thank you for the offer!  Either way... I'm happy to continue contributions through PRs, but if you need help or diversity to protect against [bus factor][], I'd be happy to be named a maintainer.  Be warned that my efforts tend to go in spurts, so don't count on me for consistently timely help.

(And, thank you for the quick responses to PRs and the rapid release cycle!)

[bus factor]: <https://en.wikipedia.org/wiki/Bus_factor>

lovasoa:
I added you !

lovasoa (approved):

PR #177, opened 2022-03-05
Update Union to accept custom marshmallow types

aaronstrateos:
@lovasoa  This updates the Union class in marshmallow_dataclasses so that it can accept polymorphic typing of custom marshmallow types as well as the predefined marshmallow types and primitive types.

### DIFF
--- a/marshmallow_dataclass/union_field.py
+++ b/marshmallow_dataclass/union_field.py
@@ -39,7 +39,9 @@ class Union(fields.Field):
         if value is None:
             return value
         for typ, field in self.union_fields:
-            if typ.__dict__.get('__supertype__') and typ.__dict__.get('_marshmallow_field'):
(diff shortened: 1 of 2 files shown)


marshmallow_dataclass/union_field.py
```diff
@@ -39,6 +39,10 @@ def _serialize(self, value: Any, attr: str, obj, **kwargs) -> Any:
         if value is None:
             return value
         for typ, field in self.union_fields:
+            if typ.__dict__.get("__supertype__") and typ.__dict__.get(
+                "_marshmallow_field"
+            ):
+                typ = typ.__supertype__
```
aaronstrateos, line 45: checks for `_marshmallow_field` to make sure this is only applied to marshmallow types

PR #179, opened 2022-03-14
Support NewType from Python 3.10

shadchin:


lovasoa:
Fixed in https://github.com/lovasoa/marshmallow_dataclass/commit/db7371470a4921ffcc004a1656be51ec209a8be9

PR #180, opened 2022-03-15
Fix use of typing.NewType in Python 3.10

noirbee:
https://docs.python.org/3.10/library/typing.html#typing.NewType

"Changed in version 3.10: NewType is now a class rather than a function."

Also add Python 3.10 to the list of tested versions in GitHub workflow.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


noirbee:
Sorry about the relative mess, I can't get it to work for Python3.6 without downgrading types-dataclasses, and since it's EOL anyways…

lovasoa (approved):

lovasoa:
Thanks

PR #187, opened 2022-04-22
Get dataclass declaration scope locals with inspect.currentframe()

noirbee:
inspect.stack() can be very costly, especially when called with
context=1: it has to get information about source files / lines
etc. This gets very slow when using dozens of dataclasses in some
large-ish codebases (where stacks can be relatively deep).

Also add the frame when class_schema() is called directly.


lovasoa:
Thanks for the pr

luk2302:
@noirbee @lovasoa I think this causes quite a significant memory leak due to `_internal_class_schema` being `lru_cache`-d and the stack frame therefore being cached and kept alive for a very long time. Especially if there are big locals around when calling `class_schema`. If we manually call `_internal_class_schema.cache_clear()` the memory is freed.

Issue #188, opened 2022-04-22
cyclic references break when using class_schema


noirbee:
v8.5.4 sadly broke recursive / cyclic schema support:

```python
import dataclasses
from typing import List

from marshmallow_dataclass import class_schema


@dataclasses.dataclass
class Tree:
    name: str
    children: List["Tree"]


class_schema(Tree)
```
```
[…]
  File "/home/nnoirbent/src/marshmallow_dataclass/marshmallow_dataclass/__init__.py", line 393, in _internal_class_schema
    attributes.update(
  File "/home/nnoirbent/src/marshmallow_dataclass/marshmallow_dataclass/__init__.py", line 396, in <genexpr>
    field_for_schema(
  File "/home/nnoirbent/src/marshmallow_dataclass/marshmallow_dataclass/__init__.py", line 686, in field_for_schema
    generic_field = _field_for_generic_type(typ, base_schema, typ_frame, **metadata)
  File "/home/nnoirbent/src/marshmallow_dataclass/marshmallow_dataclass/__init__.py", line 494, in _field_for_generic_type
    child_type = field_for_schema(
  File "/home/nnoirbent/src/marshmallow_dataclass/marshmallow_dataclass/__init__.py", line 718, in field_for_schema
    or _internal_class_schema(typ, base_schema, typ_frame)
  File "/home/nnoirbent/src/marshmallow_dataclass/marshmallow_dataclass/__init__.py", line 385, in _internal_class_schema
    for k, v in inspect.getmembers(clazz)
  File "/usr/lib/python3.10/inspect.py", line 447, in getmembers
    if isclass(object):
  File "/usr/lib/python3.10/inspect.py", line 197, in isclass
    return isinstance(object, type)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

This is due to how marshmallow_dataclass now uses `typing.get_type_hints()` for better handling of forward references, since fac0b6dc12b11c77094c4f7904ec4aadfb4d554f

Unfortunately the recursion guard from `lazy_class_attribute` only works when modifying the target class, i.e. with `@marshmallow_dataclass.dataclass` or `@add_schema` (which is triggered by this line: https://github.com/lovasoa/marshmallow_dataclass/blob/f4d04dee7831aae87a8c172dac9fa5634ad41b13/marshmallow_dataclass/__init__.py#L700 )

[...]

PR #189, opened 2022-04-25
Fix recursive references when using class_schema()

noirbee:
Fixes https://github.com/lovasoa/marshmallow_dataclass/issues/188

The @marshmallow_dataclass.dataclass decorator relies on a lazily
computed .Schema attribute/descriptor on the decorated class to solve
this issue. For class_schema() this would probably not be a valid
approach as the expectation is that its "clazz" argument would not be
modified.

Not my best work, but the `@lru_cache` on `_internal_class_schema()` made it hard to do otherwise: passing e.g. a stack of already processed classes breaks other tests which rely on the behaviour, since calls on different classes would have different stacks.

I'm also open to suggestions for a better name than `seen_classes` for this 

The `# noqa: F821` lines in the tests are due to https://github.com/PyCQA/pyflakes/issues/567

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -37,6 +37,7 @@ Full example::
 import collections.abc
 import dataclasses
 import inspect
+import threading
 import types
 import warnings
 from enum import EnumMeta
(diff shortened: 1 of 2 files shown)


lovasoa (approved):

lovasoa:
thanks

davidecotten:
Recent changes seem to break inheritance. Is that intended?

```
...


@dataclass
class ThingA:
    Schema: ClassVar[Type[Schema]]
    name: str

@dataclass
class ThingB(ThingA):
    Schema: ClassVar[Type[Schema]]
    short_name: str
```

```
>> ThingA.Schema
'ThingA'
```

```
>> ThingB.Schema
<class 'marshmallow.schema.ThingB'>
```

davidecotten:
@noirbee Thank you. That makes sense and I've confirmed the workaround.

Issue #190, opened 2022-04-26
AttributeError: '_thread._local' object has no attribute 'seen_classes'


mivade:
The latest release is raising the above error on invocation of `class_schema`. I'm working on putting together a minimal example to demonstrate this but in the meantime it's breaking CI for me.

mivade:
Here's an example:

```python
import asyncio
from dataclasses import dataclass
from marshmallow_dataclass import class_schema

@dataclass
class Foo:
    bar: int
    baz: float

async def main():
    loop = asyncio.get_running_loop()
    loop.run_in_executor(None, lambda: class_schema(Foo)())

    
if __name__ == "__main__":
    asyncio.run(main())
```

And the full traceback:

```
Future exception was never retrieved
future: <Future finished exception=AttributeError("'_thread._local' object has no attribute 'seen_classes'")>
Traceback (most recent call last):
  File "/Users/michael.depalatis/miniconda3/envs/inscripta-sherpa/lib/python3.9/site-packages/marshmallow_dataclass/__init__.py", line 356, in class_schema
    return _internal_class_schema(clazz, base_schema, clazz_frame)
  File "/Users/michael.depalatis/miniconda3/envs/inscripta-sherpa/lib/python3.9/site-packages/marshmallow_dataclass/__init__.py", line 367, in _internal_class_schema
    _RECURSION_GUARD.seen_classes[clazz] = clazz.__name__
AttributeError: '_thread._local' object has no attribute 'seen_classes'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/michael.depalatis/miniconda3/envs/inscripta-sherpa/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/michael.depalatis/tmp/untitled.py", line 12, in <lambda>
    loop.run_in_executor(None, lambda: class_schema(Foo)())
  File "/Users/michael.depalatis/miniconda3/envs/inscripta-sherpa/lib/python3.9/site-packages/marshmallow_dataclass/__init__.py", line 358, in class_schema
    _RECURSION_GUARD.seen_classes.clear()
AttributeError: '_thread._local' object has no attribute 'seen_classes'
```

Asyncio is probably not needed to reproduce but just multiple threads. I used it here since my application that is crashing is using `run_in_executor`.

mivade:
[...]

PR #191, opened 2022-04-27
Fix thread-local recursion guard initialization

noirbee:
Thread-local variables must be initialized at runtime (i.e. in their
own thread), not globally at import time.

Fixes https://github.com/lovasoa/marshmallow_dataclass/issues/190


lovasoa:
What would be the problem with a simple global variable ? Even in a multithreaded context, the GIL should be enough for it to just work naturally, no ?

lovasoa (approved):

mivade:
Hmm, did this get merged incorrectly?

image

PR #202, opened 2022-07-02
Add UnionType workaround

Vovetta:
Since UnionType from Python 3.10 currently unsupported in `typing_inspect` added workaround to use `type | type` syntax in dataclasses.


dairiki:
As of 0.8.0, `typing_inspect` supports the PEP 604 (A | B) syntax.
(And, as of release 8.5.11, we now pin `typing_inspect>=0.8.0`.)

So I think this is fixed.  Closing.

PR #204, opened 2022-08-04
added missing dataclasses arguments kw_only, slots, match_args

lorenzosciuto:

PR #205, opened 2022-08-11
with_args, kw_only, slots support in >=3.10

crutcher:
Add (new in 3.10) arguments to `@dataclass`

Similar to #204 , with sys.version_info guarding.

I tried using a common dispatch handler, but I think custom mypy handling of this class was breaking dispatch, and the forward reference tests were failing. Open to suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # marshmallow\_dataclass change log
 
+## v8.7.1 (2024-09-12)
+
+ - Relax dependency pins for `typeguard` and `typing-inspect`. ([#273], [#272])
+
+[#272]: https://github.com/lovasoa/marshmallow_dataclass/issues/272
+[#273]: https://github.com/lovasoa/marshmallow_dataclass/pull/273
+
 ## v8.7.0 (2024-06-23)
 
  - Drop support for python 3.6 and 3.7. Test under python 3.12. ([#263])
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 from setuptools import find_packages, setup
 
-VERSION = "8.7.0"
+VERSION = "8.7.1"
 
 CLASSIFIERS = [
     "Development Status :: 4 - Beta",


marshmallow_dataclass/__init__.py
```diff
@@ -81,88 +82,192 @@ class User:
...
-    order: bool = False,
-    unsafe_hash: bool = False,
-    frozen: bool = False,
-    base_schema: Optional[Type[marshmallow.Schema]] = None,
-    cls_frame: Optional[types.FrameType] = None,
-) -> Union[Type[_U], Callable[[Type[_U]], Type[_U]]]:
-    """
-    This decorator does the same as dataclasses.dataclass, but also applies :func:`add_schema`.
-    It adds a `.Schema` attribute to the class object
+    # _cls should never be specified by keyword, so start it with an
```
dairiki, line 121: Apparently, mypy recognizes parameters whose names start with a double-underscore as positional-only.
(This seems only to be documented in [the mypy cheatsheet](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html?highlight=positional-only#functions).  But the [typeshed stub] for `dataclasses.dataclass` uses the syntax.)

Anyway...  we should probably change `_cls` to `__cls`.

[typeshed stub]: https://github.com/python/typeshed/blob/95444479acf928c0c91cb25e0c6b5a2f8ade25e7/stdlib/dataclasses.pyi#L58

marshmallow_dataclass/__init__.py
```diff
@@ -81,88 +82,192 @@ class User:
...
 _RECURSION_GUARD = threading.local()
 
+if sys.version_info >= (3, 10):
+    # match_args, kw_only, slots
+
+    @overload
+    def dataclass(
+        _cls: Type[_U],
+        *,
+        repr: bool = True,
```
dairiki, line 92: While were messing around here, we should fix the typing of our `dataclass` to match [typeshed’s typing for `dataclasses.dataclass`](https://github.com/python/typeshed/blob/95444479acf928c0c91cb25e0c6b5a2f8ade25e7/stdlib/dataclasses.pyi#L55)

That allows for
```py
@dataclass
class A:
    ...
```

```py
@dataclass()
class B:
    ...
```

and

```py
@dataclass(**kwargs)
class C:
    ...
```

But does not allow for providing the class and keyword params all in one go. I.e. the following is *not supported* by `dataclasses.dataclass`:
```py
class D:
    ...

D = dataclass(D, frozen=True)
```

marshmallow_dataclass/__init__.py
```diff
@@ -81,88 +82,192 @@ class User:
...
-    cls_frame: Optional[types.FrameType] = None,
-) -> Union[Type[_U], Callable[[Type[_U]], Type[_U]]]:
-    """
-    This decorator does the same as dataclasses.dataclass, but also applies :func:`add_schema`.
-    It adds a `.Schema` attribute to the class object
+    # _cls should never be specified by keyword, so start it with an
+    # underscore.  The presence of _cls is used to detect if this
+    # decorator is being called with parameters or not.
+    def dataclass(
+        _cls: Type[_U] = None,
```
dairiki, line 125: Implicit optional parameter types are deprecated[^1].  Should fix to `__cls: Optional[Type[_U]] = None` while we're in here.  (Though there are admittedly a lot of other places in our code where we rely on implicit optional...)

[^1]: https://mypy-lang.blogspot.com/2022/09/mypy-0981-released.html

antoine-chopin:
Hello, I'm looking to use  `kw_only` with the decorator `marshmallow_dataclass`, this is a very old PR and the related issue is [closed](https://github.com/lovasoa/marshmallow_dataclass/pull/204), do you intend to add support for this?

kmc6123:
> Hello, I'm looking to use  `kw_only` with the decorator `marshmallow_dataclass`, this is a very old PR, do you intend to add support for this? Maybe it's already supported?

I made a new PR, very similar to this. Hoping to get this revived 🤞

PR #209, opened 2022-08-30
fix(ci): Don't install the library editable

vit-zikmund:
[Using the editable mode breaks path-finding capabilities of mypy](https://github.com/lovasoa/marshmallow_dataclass/pull/207#issuecomment-1222570537), which then fails for PRs.

Let's fix this so the CI checks work.


dairiki:
Running tests under an editable install should be fixed as of #211. Closing.

PR #211, opened 2022-09-23
A couple of fixes to make CI tests more resilient

dairiki:
(This PR in an alternative to #209 and also includes #207.)

### Explicitly add pytest’s rootdir to ~~MYPY~~PYTHONPATH in our mypy tests.

This helps mypy find marshmallow_dataclass when it is installed in editable mode.  (Ref #209, #207.)

### Do not fail tests upon warnings from external dependencies

We were configuring pytest with `-W error`.  This means tests failed if any warning was emitted by any dependency.
(E.g. `marshmallow<3.15` elicits a warning from its use of `distutils`.)

Here we update the warning filter so that we only fail on warnings from `marshmallow_dataclass` or our tests.

Other warnings will still be reported by `pytest` but will not cause our tests to fail.

----
## Other Changes

In order to get the CI tests to pass, we also cherry-pick PR #207 from @vit-zikmund

To get pre-commit checks to pass, we update the type of the second parameter to `Union._serialize`
to match [changes made](https://github.com/marshmallow-code/marshmallow/commit/2de97bc97891f88a56653fd583fe3635830afa2a) to `Field._serialize` typing in marshmallow 3.18.0.
(This keeps the pre-commit mypy check from failing due to "Liskov substitution principle" violation.)

Finally, we update the --target-version config for black (though I'm not sure that this makes any practical difference.)

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -48,6 +48,7 @@ from typing import (
     Dict,
     List,
     Mapping,
+    NewType as typing_NewType,
     Optional,
     Set,
     Tuple,
@@ -805,12 +806,8 @@ def NewType(
     marshmallow.exceptions.ValidationError: {'mail': ['Not a valid email address.']}
     """
 
-    def new_type(x: _U):
-        return x
-
-    new_type.__name__ = name
     # noinspection PyTypeHints
-    new_type.__supertype__ = typ  # type: ignore
+    new_type = typing_NewType(name, typ)  # type: ignore
     # noinspection PyTypeHints
     new_type._marshmallow_field = field  # type: ignore
     # noinspection PyTypeHints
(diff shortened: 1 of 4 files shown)


lovasoa:
Wow, great ! But it looks like the tests still fail in the ci

dairiki:
> Wow, great ! But it looks like the tests still fail in the ci

Of course they do! LOL

I've fixed things again so that they now pass.  I had to include the changes from #207 to get that to happen.  I've updated my original description of this PR to match what's currently in it.

@lovasoa: Do you want to continue support for python 3.6 (EOL 2021-12), or can that be cleaned out?

lovasoa (approved):

lovasoa:
Thank you very much @dairiki and  @vit-zikmund :)

lovasoa:
For python 3.6, it looks like there still are hundreds of daily downloads of marshmallow-dataclass with this python version: https://pypistats.org/packages/marshmallow-dataclass

But eventually, when python 3.6 becomes too much of a burden to support, and there are fewer daily downloads, we'll drop support, yes :)

Issue #210, opened 2022-09-21
Basic support for partial=True


hhalaby:
This relates to #169 but instead of suggesting to return a different object type depending on partial, I suggest to simply return a dataclass with missing fields much like marshmallow does for its dicts. Otherwise the partial argument cannot be used. 

If someone is specifically setting `partial=True`, it means they are aware of the consequences.

dairiki:
But (if I understand correctly) a "dataclass with missing fields" is also a different object type than the original dataclass.

E.g.
```py
@dataclass
class C:
    x: int
    y: int

@dataclass
class PartialC:
    x: int

c = C(1,2)
pc = PartialC(1)

isinstance(c, PartialC) # => False
isinstance(pc, C) # => False
```

I.e. dataclasses don't support "missing fields".  A dataclass without a particular set of fields is a whole new type.
If `load` (with `partial=True`) were to return a whole bunch of different (dynamically created) types based on which input fields were missing, I think the behavior would be quite confusing in practice.

hhalaby:
@dairiki Thank you for your explanation. I think I had misunderstood at first the implications of what I suggested. More so for my specific use case of wanting to create unit tests for specific parts of a dataclass. No need to load a partial dataclass, all I have to do is load a complete one.

Issue #169, opened 2021-12-14
No support for handling partial=True schemas


iburakov:
Currently there seems to be no out-of-the-box way to handle partial schemas:
```python
from marshmallow_dataclass import dataclass

@dataclass                                                                                                                             
class Foo:
    x: int
    y: str

Foo.Schema(partial=True).load({"x": 3, "y": "bar"})  # => Foo(x=3, y='bar')
Foo.Schema(partial=True).load({"x": 3})  # => TypeError: __init__() missing 1 required positional argument: 'y' 
```

The abovementioned error is raised when schema's overridden .load() function tries to make a dataclass instance out of marshmallow's dict on line 644:
https://github.com/lovasoa/marshmallow_dataclass/blob/fa6c28980ccfe45742cdc9430bfb3b737690935f/marshmallow_dataclass/__init__.py#L638-L644

Would it be useful to change this method to return marshmallow's .load() result without trying to unpack it into a dataclass for partial=True schemas? In my opinion, making a dataclass out of partial dict doesn't make much sense.

As a workaround, we're overriding the .load() method again in our codebase to achieve such behavior:
```python
from marshmallow import Schema
from typing import Mapping

class FooSchema(Foo.Schema):
    def load(self, data: Mapping, *, many: bool = None, partial: bool = None, **kwargs):
        """ Customized "load" method to skip unpacking to dataclass for partial=True schemas """
        if partial is True or self.partial is True:
            return Schema.load(self, data, many=many, partial=partial, **kwargs)
        else:
            return super().load(data, many=many, partial=partial, **kwargs)

Foo.Schema = FooSchema
Foo.Schema().load({"x": 3, "y": "bar"})  # => Foo(x=3, y='bar'), as in the previous example
Foo.Schema().load({"x": 3})  # => marshmallow's ValidationError: {'y': ['Missing data for required field.']} - schema isn't partial
Foo.Schema(partial=True).load({"x": 3})  # => {'x': 3} - schema is partial now, raw dict is loaded and returned, as anticipated
```

Is it reasonable to incorporate such support for partial=True schemas into the library itself?

lovasoa:
I think the current behavior is reasonable, and returning a different object type when partial=True instead of raising an error would be surprising

iburakov:
You can make it an opt-in behavior if you think it's too unintuitive.

[...]

PR #213, opened 2022-09-23
Add .load_to_mapping method to base schema

dairiki:
Following up on [this suggestion] made in a comment on #169 (see also #210), this provides a new schema method, `.load_to_mapping`, that essentially does a raw `marshmallow.Schema.load` without converting any of the deserialized values (either top-level or nested) to dataclasses.  The main point of this is to allow for deserializing partial data (with `partial=True`).

This PR is provided for discussion, as to:
- whether it's worthwhile
- whether there's a better name for the new method
- whether there's a better way to implement the new method
- ...

[this suggestion]: https://github.com/lovasoa/marshmallow_dataclass/issues/169#issuecomment-1256298997

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -40,6 +40,7 @@ import inspect
 import threading
 import types
 import warnings
+from contextlib import contextmanager
 from enum import EnumMeta
 from functools import lru_cache, partial
 from typing import (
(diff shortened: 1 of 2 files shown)


lovasoa:
I'm ok with that ! Maybe `load_to_dict` is clearer than `load_to_mapping` ? Doesn't it always return a dict ?

dairiki:
> I'm ok with that ! Maybe `load_to_dict` is clearer than `load_to_mapping` ? Doesn't it always return a dict ?

Yes, that's probably better.

After some digging through the marshmallow source code it appears to always return dict (or OrderedDict) (or a list of those if `many` is set).  I will rename it.

Issue #225, opened 2022-12-22
Add native marshmallow.fields.Enum suport


uvsmtid:
`Enum` support was added by `marshmallow` v3.18.0:
https://github.com/marshmallow-code/marshmallow/commit/fa0ee03e82d2b73dc00bdeb75b266aba99be8d0b

`EnumField` from `marshmallow-enum` becomes unnecessary:
https://github.com/justanr/marshmallow_enum/issues/51

In fact, native support is necessary to generate OAS spec from marshmallow specs using `apispec`.
This line does not support `EnumField` from `marshmallow-enum`, it wants `Enum` from `marshmallow`:
https://github.com/marshmallow-code/apispec/blob/bd0d12f7b932369f718d1b222a68f1ddff399953/src/apispec/ext/marshmallow/field_converter.py#L512

Suporting `Enum` makes all pieces work togeter.

dairiki:
Done via #227.  Released, just now, in marshmallow-dataclass 8.5.11.

PR #227, opened 2023-01-07
Replace marshmallow-enum with native marshmallow.fields.Enum

otonnesen:
Addresses https://github.com/lovasoa/marshmallow_dataclass/issues/225

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,11 +56,10 @@ pip3 install marshmallow-dataclass
 
 You may optionally install the following extras:
 
-- `enum` : for translating python enums to [marshmallow-enum](https://github.com/justanr/marshmallow_enum).
 - `union` : for translating python [`Union` types](https://docs.python.org/3/library/typing.html#typing.Union) to union fields.
 
 ```shell
-pip3 install "marshmallow-dataclass[enum,union]"
+pip3 install "marshmallow-dataclass[union]"
 ```
 
 ### marshmallow 2 support
--- a/README.md
+++ b/README.md
@@ -56,10 +56,11 @@ pip3 install marshmallow-dataclass
 
 You may optionally install the following extras:
 
+- `enum` : enum support for marshmallow versions <3.18 [marshmallow-enum](https://github.com/justanr/marshmallow_enum).
 - `union` : for translating python [`Union` types](https://docs.python.org/3/library/typing.html#typing.Union) to union fields.
 
 ```shell
-pip3 install "marshmallow-dataclass[union]"
+pip3 install "marshmallow-dataclass[enum,union]"
 ```
 
 ### marshmallow 2 support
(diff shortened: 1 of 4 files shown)


dairiki:
The CI test is failing under python 3.6.

Support for python 3.6  was dropped in `marshmallow` 3.15 (and Enum support was only added in `marshmallow` 3.18.)

Two options (perhaps there are others) appear to be:

- Drop python 3.6 support in `marshmallow_dataclass`.  The last time I asked, @lovasoa was against doing that.  He'll have to say whether it's time to do that yet.
- Add extra code/smarts so as to be able to use `marshmallow_enum.EnumField` under python 3.6 and `marshmallow.fields.Enum` otherwise.

mivade:
Python 3.6 reached its [end of life](https://endoflife.date/python) over a year ago now. I think it would make sense to drop support for it.

otonnesen:
In the interest of keeping this diff self-contained, I'll just add the extra code to keep python 3.6 support as you mentioned. I can open another PR to discuss dropping 3.6 support as well if we like.

dairiki:
> Python 3.6 reached its [end of life](https://endoflife.date/python) over a year ago now. I think it would make sense to drop support for it.

I do not disagree, however, the [last time this came up](https://github.com/lovasoa/marshmallow_dataclass/pull/211#issuecomment-1256615987), @lovasoa wanted to maintain 3.6 support, citing the number of weekly downloads.  Since then, download numbers have dropped a bit, but not substantially.

image

One option might be to bump our minor version number, dropping support for py 3.6 in the new version, and maintain two branches (an 8.5.x that supports py3.6, and an 8.6.x which does not).

But, as long as it's not too painful to maintain 3.6 support in the current codebase, it's probably cleaner and less work just to do so.

dairiki:
@otonnesen Would it be worthwhile (and actually work) to make the `[enum]` extra conditional on python version?

E.g., in setup.py:

```py
EXTRAS_REQUIRE = {
    "enum": [
        "marshmallow-enum; python_version<'3.7'",
        "marshmallow>=3.18; python_version>='3.7'",
    ],
    ...
}
```

dairiki:
> @otonnesen Would it be worthwhile (and actually work) to make the `[enum]` extra conditional on python version?

Never mind.  I see you just did that! :laughing:

otonnesen:
Yeah just tested on 3.6 and 3.10 and it does appear to work as I'd expect

dairiki:
Will wait for further comments for a couple of days before merging, but LGTM.  Thank you @otonnesen!

lovasoa (approved):

lovasoa:
Good for me!

dairiki:
Released, just now, in 8.5.11.  Thank you, @otonnesen!

Issue #166, opened 2021-10-12
Union no longer passes metadata to component fields


dcchut:
The following snippet runs successfully using `marshmallow_dataclass[enum,union]==8.5.2`:
```python
import enum
from typing import Union

import marshmallow_dataclass
from dataclasses import dataclass, field


class Fruit(enum.Enum):
    apple = "Apple"
    banana = "Banana"
    tomato = "Tomato"

@dataclass
class Punnet:
    fruit: Union[Fruit, dict] = field(metadata=dict(by_value=True))

schema = marshmallow_dataclass.class_schema(Punnet)()
assert schema.load({"fruit": "Tomato"}) == Punnet(fruit=Fruit.tomato)
```
As of 8.5.3 this no longer runs, instead raising an error:
```
marshmallow.exceptions.ValidationError: {'fruit': [ValidationError('Invalid enum member Tomato',), ValidationError('Not a valid mapping type.',)]}
```

I believe this was caused by the change in #162 (previously the `by_value` flag was being passed to the enum field).

gnedivad:
Can we update `_field_for_generic_type` to pass through the `metadata` [1] i.e. update that line to be
```python
metadata={**metadata, "required": True}
```

[1] https://github.com/lovasoa/marshmallow_dataclass/blob/6b6dc51c528e7c859a04f80c8af97c9cc1850bdb/marshmallow_dataclass/__init__.py#L582

dairiki:
> Can we update `_field_for_generic_type` to pass through the `metadata` [1] i.e. update that line to be
> 
> ```python
> metadata={**metadata, "required": True}
> ```

This suggestion here is to pass (all or most) of the metadata specified for the `Union` field on to the component fields of the union.
That feels fishy to me, as there's no way to control which of the subfields receives which metadata.  What metadata is meant for the Union field, what is meant for the enum subfield, and what is meant for the dict subfield?  It seems unlikely that we really want all three of those to receive the same metadata, at least in the general case.
(Here we're passing `by_value=True` on to the `Mapping` field for the `dict` as well — that's harmless in this particular case, but not always.)

Issue #167 is related, having to do with how to pass metadata to subfields of a `List` field.

----

Here's one way around the issue, using `marshmallow_dataclass.NewType` to specify metadata for the enum subfield.

```py
import enum
from typing import Union

from dataclasses import dataclass
from marshmallow_dataclass import class_schema, NewType

class Fruit(enum.Enum):
[...]

PR #231, opened 2023-01-19
add support for union with metadata #166

AnkurDedania:


dairiki:
Thank you for the PR.

I’m not convinced that this is the correct solution to the issue.
When constructing a schema for:
```py
@dataclass
class X
    y: Union[SomeType, OtherType] = field(metadata=some_metadata)
```
we need to construct three schema fields: one for SomeType, one for OtherType, and finally one for the Union.  In the general case, it seems unreasonable to assume that we want to pass exactly the same custom metadata to all three of those fields. 

I've just left [a comment](https://github.com/lovasoa/marshmallow_dataclass/issues/166#issuecomment-1397787705) on #166 with more on this.

PR #235, opened 2023-03-15
Fixes for typeguard>=3.0.0rc2 (and mypy 1.1.1)

### LINKED ISSUE #234: Issues with typeguard package v3
`typeguard` version 3 does not have the `argname` argument to `check_type()`, raising an exception in `marshmallow_dataclass`.

Moreover, the version of `typeguard` is not constrained in `setup.py`, causing a silent upgrade of `typeguard` in projects that use `marshmallow_dataclass` even when those projects fully constrain the version of `marshmallow_dataclass`, which breaks those projects in a surprising way. It caused me an hour to debug one of my projects. The version of `typeguard` should have at least the major version number contrained.

dairiki:
Fixes #234 

Also fixes issues with our mypy plugin with mypy>=1.1.1.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
 - repo: https://github.com/asottile/pyupgrade
-  rev: v3.2.3
+  rev: v3.3.1
   hooks:
   - id: pyupgrade
     args: ["--py36-plus"]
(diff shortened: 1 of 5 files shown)


dkmiller (approved):
LGTM, thanks for the quick fix!

dairiki:
Released in 8.5.12.

PR #239, opened 2023-04-19
dataclass_transform: use regular dataclasses.field for field_specifiers

noirbee:
cf https://docs.python.org/3/library/typing.html#typing.dataclass_transform: this is needed so that type checkers properly handle attributes using field(). Without this, any attribute using `= field(...)` is treated as if it were a default value by mypy, which makes it impossible to use its other arguments. This is even more problematic for marshmallow-dataclass since it makes it impossible to pass arguments to marshmallow using `metadata=...`

(note that this only affects users of mypy >= 1.1, since this is the release in which `@dataclass_transform` support was added)

IMHO this should probably be the default value in Python's `typing` module / `typing_extensions` but changing it in marshmallow-dataclass is probably easier/faster to get it to work ASAP :)

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -71,8 +71,12 @@ from marshmallow_dataclass.lazy_class_attribute import lazy_class_attribute
 
 if sys.version_info >= (3, 11):
     from typing import dataclass_transform
-else:
+elif sys.version_info >= (3, 7):
     from typing_extensions import dataclass_transform
+else:
+    # @dataclass_transform() only helps us with mypy>=1.1 which is only available for python>=3.7
+    def dataclass_transform(**kwargs):
+        return lambda cls: cls
 
 
 __all__ = ["dataclass", "add_schema", "class_schema", "field_for_schema", "NewType"]
@@ -122,7 +126,7 @@ def dataclass(
 # _cls should never be specified by keyword, so start it with an
 # underscore.  The presence of _cls is used to detect if this
 # decorator is being called with parameters or not.
-@dataclass_transform()
+@dataclass_transform(field_specifiers=(dataclasses.Field, dataclasses.field))
 def dataclass(
     _cls: Optional[Type[_U]] = None,
     *,
--- a/tests/test_mypy.yml
+++ b/tests/test_mypy.yml
@@ -65,3 +65,18 @@
     reveal_type(website.email)  # N: Revealed type is "builtins.str"
 
     Website(url=42, email="user@email.com")  # E: Argument "url" to "Website" has incompatible type "int"; expected "str"  [arg-type]
+- case: dataclasses_field_not_a_default
+  mypy_config: |
+    follow_imports = silent
+    plugins = marshmallow_dataclass.mypy
+    show_error_codes = true
+  env:
+    - PYTHONPATH=.
+  main: |
+    from dataclasses import field
+    from marshmallow_dataclass import dataclass
+
+    @dataclass
+    class User:
+        id: int = field(metadata={"required": True})
+        name: str


noirbee:
The tests for Python3.6 fail because it uses typing-extensions v4.1.1, where `dataclass_transform` has a `field_descriptors` argument. In v4.2.0 `dataclass_transform` was updated to rename it `field_specifiers` instead, while also dropping Python3.6 support ( https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-420-april-17-2022 )

I'm trying to support it by inspecting the function's signature, but at this point I wonder if it's worth it vs bumping typing-extension to >4.2 and dropping support for python3.6, which has now been EOLed for over a year.

dairiki:
> I'm trying to support it by inspecting the function's signature, but at this point I wonder if it's worth it vs bumping typing-extension to >=4.2 and dropping support for python3.6, which has now been EOLed for over a year.

I'd be all for dropping python3.6 support.  However, the choice on whether to do that is up to the project owner (@lovasoa), who, the last time the issue was brought up, still thought it was worth keeping python 3.6 support.  (It some months later now, though, so maybe his answer has changed?)

In any case, for now, how about just something like:
```py
if sys.version_info >= (3, 11):
    from typing import dataclass_transform
elif sys.version_info >= (3,7):
    from typing_extensions import dataclass_transform
else:
    # @dataclass_transform() only helps us with mypy>=1.1 which is only runs under python>=3.7
    def dataclass_transform(**kwargs):
        return lambda x: x
```
    
```

File: marshmallow_dataclass/__init__.py [hunk 1]
```diff
@@ -122,7 +122,7 @@ def dataclass(
 # _cls should never be specified by keyword, so start it with an
 # underscore.  The presence of _cls is used to detect if this
 # decorator is being called with parameters or not.
-@dataclass_transform()
+@dataclass_transform(field_specifiers=(dataclasses.field,))
```
dairiki, line 125: According to [PEP 681], this should be
```
field_specifiers=(dataclasses,Field, dataclasses.field)
```

[pep 681]: https://peps.python.org/pep-0681/#dataclass-transform-parameters:~:text=The%20standard%20dataclass%20behavior%20supports%20only%20one%20type%20of%20field%20specifier%20called%20Field%20plus%20a%20helper%20function%20(field)%20that%20instantiates%20this%20class%2C%20so%20if%20we%20were%20describing%20the%20stdlib%20dataclass%20behavior%2C%20we%20would%20provide%20the%20tuple%20argument%20(dataclasses.Field%2C%20dataclasses.field).

lovasoa:
Last time we talked about it, we checked the stats and it seemed like the was still a non negligible amount of downloads from 3.6 

Looks like it is almost zero now:
https://pypistats.org/packages/marshmallow-dataclass

So I'm okay with dropping support. We just need to make the next release a new major version.

dairiki:
> So I'm okay with dropping support. We just need to make the next release a new major version.

Since (I think) the workaround to support py36 in this PR is simple and clean enough, I say we go for that.

----

Also, I would think that when/if we do drop py36 support, bumping the minor version is sufficient to cover that.
(There is arguably no backward-incompatible API change associated with dropping support of old pythons.)
In any case, so long as we update [Requires-Python](https://peps.python.org/pep-0345/#requires-python) appropriately, there is no breakage for users of python 3.6.  For them, pip will continue, by default, to install the highest available version that claims to support python 3.6.

marshmallow_dataclass/__init__.py
noirbee: Good catch !

noirbee:
Updated the PR with @dairiki 's suggestions; using a dummy `@dataclass_transform()` works much better / is simpler than the approach I first tried (using `inspect.signature()` / and generating a kwargs dict on the fly), thanks !

lovasoa (approved):
@dairiki , feel free to merge :)

dairiki:
Merged and released as 8.5.13.

Thank you, @noirbee !

PR #241, opened 2023-05-03
Specify build backend

EpicWink:
Specify project package build backend (setuptools)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
+build-backend = "setuptools.build_meta"
+
 [tool.black]
 line-length = 88
 target-version = ['py36', 'py37', 'py38', 'py39', 'py310', 'py310']


dairiki:
> Downsides: `python setup.py` stops working (desired), requires pip v19 (from 2019)

I don't have strong feelings on this either way, other than that I think it is confusing to have a `setup.py` if it doesn't work.

I think the project metadata could be moved to `setup.cfg` (and then `setup.py` deleted) while still remaining buildable under Python 3.6.  But that's probably not worth it, since the longer-term goal is to move the metadata to `pyproject.toml` (#242) which, it appears, can not be done while maintaining Python 3.6 compatibility.

So, I guess, I would rather hold off on this until we drop Python 3.6 support (which is "imminent"), at which point we can fold this PR into #242.

EpicWink:
> I don't have strong feelings on this either way, other than that I think it is confusing to have a setup.py if it doesn't work.

Invoking setup.py as a script will stop working (because setuptools may not be installed), which is fine as at some point setuptools will drop support for this. setup.py can still be used to specify project configuration.

Issue #60, opened 2019-12-27
Allowing of init=False fields to appear in the schema


ori-levi:
Can you please an argument to the class_schema method to allow appear fields with init=False in the schema?

This option will be very useful when using inheritance and default arguments.

And what do you think about adding to the schema class a suffix of 'Schema'

ori-levi:
I might have a solution for this issues, I will implement it and send a pull request :)

lovasoa:
I think we should keep this open to discuss whether we want to add this behavior or not.

Link to PR: #61

lovasoa:
Can you describe your use case and what you are trying to achieve ?

ori-levi:
Sure,
In my workplace, we use to represent some entities in our database.
Our database (MongoDB) generate id, and we want to store the id in our models, load and dump it,
and if we create an instance of the model, we don't want to pass the id.

So it looks like this:

`@dataclass
class _BaseModel:
    schema: ClassVar[Schema]
    id: Optional[ObjectId] = field(
        init=False,
        default=None,
        metadata={'marshmallow_field': ObjectIdField(data_key='_id',
                                                     required=False,
                                                     allow_none=True,
                                                     as_str=True)}
    )

@dataclass
class Model(_BaseModel):
    port: int
    host: str
`

lovasoa:
I feel like it would make more sense and be easier to just make your id field `init=True`, and optionally create a separate class method if you want to create instances without providing an id. Wouldn't it ?

ori-levi:
This is just a simple example, we have a lot of other fields with `initial=False`

lovasoa:
And what would be the issue with creating a class methods for creating instances of them ?

ori-levi:
It's not just the Id property, here is another example
```
class SettingType:
    SYSTEM = 'System'
    ALERTS = 'Alerts'

@dataclass
class _BaseModel:
    schema: ClassVar[Schema]
    id: Optional[ObjectId] = field(
        init=False,
        default=None,
        metadata={'marshmallow_field': ObjectIdField(data_key='_id',
                                                     required=False,
                                                     allow_none=True,
                                                     as_str=True)}
    )
    type: SettingType = field(init=False, default=None, metadata={'marshmallow_field': EnumField(SettingType)})

    def __init_subclasses__(cls, **kwargs):
        super().__init_subclasses__(**kwargs)
        cls.type = kwargs['type']

@dataclass
class Model(_BaseModel,
[...]

PR #246, opened 2023-09-15
Add Meta option to include non init-ed fields

huwcbjones:
Updates #60

This adds the ability to include non init-ed fields into the schema by using a new Meta option , `include_non_init`.

I don't think it's the best way to address #60, but it is a very small code change to get the functionality working.

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -297,6 +297,9 @@ def class_schema(
     >>> person
     Person(name='Anonymous', friends=[Person(name='Roger Boucher', friends=[])])
 
+    Marking dataclass fields as non-initialized (``init=False``), by default, will result in those
+    fields from being exluded in the schema. To override this behaviour, set the ``Meta`` option
+    ``include_non_init=True``.
     >>> @dataclasses.dataclass()
     ... class C:
     ...   important: int = dataclasses.field(init=True, default=0)
@@ -309,6 +312,20 @@ def class_schema(
     ... }, unknown=marshmallow.EXCLUDE)
     >>> c
     C(important=9, unimportant=0)
+
+    >>> @dataclasses.dataclass()
+    ... class C:
+    ...   class Meta:
+    ...     include_non_init = True
+    ...   important: int = dataclasses.field(init=True, default=0)
+    ...   unimportant: int = dataclasses.field(init=False, default=0)
+    ...
+    >>> c = class_schema(C)().load({
+    ...     "important": 9, # This field will be imported
+    ...     "unimportant": 9 # This field will be imported
+    ... }, unknown=marshmallow.EXCLUDE)
+    >>> c
+    C(important=9, unimportant=9)
 
     >>> @dataclasses.dataclass
     ... class Website:
@@ -408,6 +425,9 @@ def _internal_class_schema(
         if hasattr(v, "__marshmallow_hook__") or k in MEMBERS_WHITELIST
     }
 
+    # Determine whether we should include non-init fields
+    include_non_init = getattr(getattr(clazz, "Meta", None), "include_non_init", False)
+
     # Update the schema members to contain marshmallow fields instead of dataclass fields
     type_hints = get_type_hints(
         clazz, localns=clazz_frame.f_locals if clazz_frame else None
@@ -424,7 +444,7 @@ def _internal_class_schema(
             ),
         )
         for field in fields
-        if field.init
+        if field.init or include_non_init
     )
 
     schema_class = type(clazz.__name__, (_base_schema(clazz, base_schema),), attributes)
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python_version: ["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.9"]
+        python_version: ["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.10"]
         include:
           - os: "ubuntu-20.04"
             python_version: "3.6"
--- a/tests/test_class_schema.py
+++ b/tests/test_class_schema.py
@@ -434,6 +434,29 @@ class TestClassSchema(unittest.TestCase):
             {"first": {"second": {"first": None}}},
         )
 
+    def test_init_fields(self):
+        @dataclasses.dataclass
+        class NoMeta:
+            no_init: str = dataclasses.field(init=False)
+
+        @dataclasses.dataclass
+        class NoInit:
+            class Meta:
+                pass
+
+            no_init: str = dataclasses.field(init=False)
+
+        @dataclasses.dataclass
+        class Init:
+            class Meta:
+                include_non_init = True
+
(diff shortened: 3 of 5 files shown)


marshmallow_dataclass/__init__.py
```diff
@@ -408,6 +408,9 @@ def _internal_class_schema(
         if hasattr(v, "__marshmallow_hook__") or k in MEMBERS_WHITELIST
     }
 
+    # Determine whether we should include non-init fields
+    include_non_init = getattr(getattr(clazz, "Meta", None), "include_non_init", False)
```
lovasoa, line 412: This would need to be documented...

lovasoa (changes requested):

marshmallow_dataclass/__init__.py
huwcbjones: Are you happy with the interface? If so, will go ahead and add some documentation

lovasoa: The interface is fine by me. But the tests are failing

huwcbjones:
@lovasoa the tests appear to also be broken on master due to bitrot, have fixed those.
Also added documentation

huwcbjones:
Have pushed a better fix for the mypy test failure

lovasoa:
Thanks for the cleanup !

huwcbjones:
Last failing test seems to be some ordering thing that I have no idea how to fix 😬

dairiki:
I've just pushed a commit that refactors `TestFieldForSchema.assertFieldsEqual` to make it somewhat more robust.

It could probably use a little more cleanup and review.  (E.g. to make sure I haven't inadvertently made the check too lenient.)

lovasoa:
The problem was with the order of union fields. But are you sure we don't care about it? If there is a logic of "first field that works", then it won't behave the same. 
I think the actual problem is in the code, not the tests.

dairiki:
> The problem was with the order of union fields. But are you sure we don't care about it? If there is a logic of "first field that works", then it won't behave the same.

You're right, of course.  Crap.

(Though I still think the rest of the refactor to `assertFieldsEqual` is an improvement over the previous implementation, which mostly relied on comparing reprs.)

lovasoa:
I was writing a comment about comparing the values to themselves, but I see you realized it yourself. But the point is: it's much more complicated and error prone that way. Can we go back to just the reprs ? It was very simple and worked well.

dairiki:
Here's the issue with the Union field ordering. (I think.)

It has to do with the fact that `typing` caches some types.

```
>>> from typing import Optional, Union
>>> Optional[Union[int, str]] is Optional[Union[str, int]]
True

# Even though...
>>> Union[int, str] is Union[str, int]
False
```

The failure of our [test_optional_multiple_types] is exercised by [this change] in `pytest-mypy-plugins` 1.11.0.

[this change]: https://github.com/typeddjango/pytest-mypy-plugins/commit/2459905e935f55846a5936ccbccbe4ae711ab5a0#diff-6e9a599f2138b2578adb39006a71c88eb253ffac37167235fa00e737a5469542L90-R90

The cause of all this is that, while our `Union` field is order-dependent, union types semantically do not depend on the order of their arguments.  (`Union[str, int]` and `Union[int, str]` refer to the same type.)

I'm not sure what the fix is.

lovasoa:
Wow, that is strange ! If there is really no way to extract the original order of the union members, then we don't have any choice but to special-case our tests to allow any order for the members of `union_fields`

dairiki:
> Can we go back to just the reprs ?

Okay, I've reverted my changes.

I've left my reversion of e8e29b9a45ad76d7894910b9350cb025071c28b8, and partial reversion of cbad82bdce3d8d7605f2767c050ca1f539d7d5b5 since both of those attempted to fix tests by making them insensitive to Union field order.

lovasoa:
Ok, thanks to both of you, I'm going to release a new version

dairiki:
> Wow, that is strange ! If there is really no way to extract the original order of the union members, then we don't have any choice but to special-case our tests to allow any order for the members of `union_fields`

But the tests <del>are</del>were correct!  (They are correctly revealing a problem.)  You correctly pointed out that the behavior of our `Union` field depends on the ordering.  Our support for union types is broken if union types don't preserve argument order.

----

Here's a clearer explanation of what's going on...

`Typing` maintains a cache of constructed types.  The cache is (essentially) keyed on the arguments to the type.

`Union[str, int]` and `Union[int, str]` have different arguments (`(str, int) != (int, str)`), so they evaluate to different instances.

But, even though they are different instances, `Union[str, int] == Union[int, str]`.  Union types  [don't consider argument order when computing equality](https://github.com/python/cpython/blob/e57ecf6bbc59f999d27b125ea51b042c24a07bd9/Lib/typing.py#L1531-L1538).

So `Union[Union[int, str]]` will return the same (cached) instance as `Union[Union[str, int]]`, because the arguments to the outer `Union` compare equal (`(Union[int, str],) == (Union[str, int],)`).  (`Optional[Union[int, str]]` is equivalent to `Union[Union[int, str], type(None)]`.)

----
Brainstorming...

Maybe we could implement our own custom union type that would bypass `typing`s type cache and so preserve the argument ordering but otherwise work the same as `typing.Union`.  (I'm not sure how easy that is to do in a way that will work with all supported python versions.)

We could maybe somehow monkeypatch `typing.Union` to make it bypass the type cache.  (Yuck.  Again, not sure how easy this is to do in a python-version-independent say.)

lovasoa:
You are right ! You can make such a patch, but personally, I'd just wait to see if anyone needs it. This theoretically can be a problem, but I'm not sure how urgent the fix is...

dairiki:
> Ok, thanks to both of you, I'm going to release a new version

@lovasoa The Union type ordering issue has nothing to do with the original point of the PR, so I don't object to making a release to get the `include_non_init` option out.

But, our handling of union types is quite broken. (And I think it always has been?)

The tests were correct in checking the order of `union_fields`.  Your last commit should be reverted in master and an issue created.

Here's a simple demo to illustrate the issue.  (Put this in `broken.py` then run `python broken.py`.)

```py
from typing import Optional, Union

# Comment out the next line and the assert will pass, otherwise it will fail
Optional[Union[str, int]]

@dataclass
class Test:
    x: Optional[Union[int, str]]

assert Test.Schema().load({"x": "42"}) == Test(x=42)
```

lovasoa:
I agree we should create an issue and document the current clunky behavior, but I don't think we should re-break the tests on master

dairiki:
> I agree we should create an issue and document the current clunky behavior, but I don't think we should re-break the tests on master


The failing tests could be marked [xfail](https://docs.python.org/3/library/unittest.html#unittest.expectedFailure)

lovasoa:
don't you think it's better to still test that the union members are all here and correct than giving up on testing union altogether ?

dairiki:
> don't you think it's better to still test that the union members are all here and correct than giving up on testing union altogether ?

We could do both.  Mark the existing failing tests as expected failures.  Add a new test (or tests) to check things that currently work that were checked by the failing tests.

I think the *xfail* tests are important as reminders of what is broken.  Since we know there's a problem, there should be a test that exercises it.

lovasoa:
Good, I'm fine with that !

Issue #247, opened 2023-09-16
Union field is broken: The order of Union arguments is indeterminate


dairiki:
Semantically, the order of the arguments of a union type is insignificant.  `Union[int, str]` and `Union[str, int]` refer to the same type.  When we ask a `typing.Union` instance for its arguments, it is under no obligation to give them to us in the same order as was used to create the instance.

Our [`Union` marshmallow field], however, does attach significance to the order of the arguments.  When de/serializing a value, it tries to do so using a field appropriate for each of the types in the union, *in order*.  E.g. for a (dataclass) field with type `Union[int str]`, we try to deserialize first as an `int`, falling back to `str`, only if the input can not be parsed as an integer.

So the whole design of our system for handling union fields is flawed.

Most of the time, it happens to work.  Python's typing does have a *type cache*. When a generic alias (e.g. a `Union`) is instantiated, a previously created instance may be returned, if one has already been created with the same arguments.

The type cache system (currently) does not ignore argument order, a fact which mostly saves us:
```py
>>> from typing import *
>>> Union[int, str] is Union[int, str]
True
>>> Union[int, str] is Union[str, int]
False
>>> Union[int, str].__args__
(<class 'int'>, <class 'str'>)
>>> Union[str, int].__args__
(<class 'str'>, <class 'int'>)
```

Equality comparison between Unions, however, does ignore the argument order:
```py
>>> Union[int, str] == Union[str, int]
True
```

When arguments to a union type are themselves unions, this can start to cause trouble.
```py
>>> Union[Union[int, str]] is Union[Union[str, int]]
True
>>> Union[Union[int, str]]
typing.Union[int, str]
>>> Union[Union[str, int]]
typing.Union[int, str]
```

Because the argument lists to the outer calls to `Union` are equal, we get the same instance back.  One of the two results will have the "wrong" argument order.

----

Here's a simple script that exercises the problem.

```py
from typing import Optional, Union
from marshmallow_dataclass import dataclass

# Comment out the next line and the assert will pass, otherwise it will fail
Optional[Union[str, int]]

@dataclass
class Test:
    x: Optional[Union[int, str]]

assert Test.Schema().load({"x": "42"}) == Test(x=42)
```

### Related

This issue surfaced and was discussed in [PR #246](https://github.com/lovasoa/marshmallow_dataclass/pull/246#issuecomment-1722291806).

PR #248, opened 2023-09-16
Do not ignore Union argument ordering in tests

dairiki:
The union argument ordering is important!

What's here:
- Revert b61cd5035b7ec0722e1ac7a361cbd9e8f6a4387a. That commit effectively hides failing tests by ignoring the argument order of union fields.
- The one test that fails because of that reversion is marked an expected failure.
- A new version of that failing test is added.  The new version does ignore the union argument ordering (and so is expected to pass).

### Related

The failing test fails due to #247.

This was discussed in [PR #246](https://github.com/lovasoa/marshmallow_dataclass/pull/246#issuecomment-1722295574).

### DIFF
--- a/tests/test_field_for_schema.py
+++ b/tests/test_field_for_schema.py
@@ -1,3 +1,4 @@
+import copy
 import inspect
 import sys
 import typing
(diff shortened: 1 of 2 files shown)


tests/test_field_for_schema.py
```diff
@@ -203,6 +201,25 @@ def test_optional_multiple_types(self):
             ),
         )
 
+    def test_optional_multiple_types_ignoring_union_field_order(self):
+        result = field_for_schema(Optional[Union[int, str]])
```
lovasoa, line 205: Maybe add a link for context here ?

```suggestion
        # see https://github.com/lovasoa/marshmallow_dataclass/pull/246#issuecomment-1722204048
        result = field_for_schema(Optional[Union[int, str]])
```

PR #252, opened 2024-01-21
add slots, match_args, kw_only to dataclass func

kmc6123:
This is similar to an outstanding PR for two reasons:
- this PR contains less code and is more explicit imo 
- the outstanding PR is about 2 years old, hoping to revive it as this feature would be helpful for me, my team, and others

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -103,6 +103,9 @@ def dataclass(
     order: bool = False,
     unsafe_hash: bool = False,
     frozen: bool = False,
+    match_args: bool = True,
+    kw_only: bool = False,
+    slots: bool = True,
     base_schema: Optional[Type[marshmallow.Schema]] = None,
     cls_frame: Optional[types.FrameType] = None,
 ) -> Type[_U]:
@@ -117,6 +120,9 @@ def dataclass(
     order: bool = False,
     unsafe_hash: bool = False,
     frozen: bool = False,
+    match_args: bool = True,
+    kw_only: bool = False,
+    slots: bool = True,
     base_schema: Optional[Type[marshmallow.Schema]] = None,
     cls_frame: Optional[types.FrameType] = None,
 ) -> Callable[[Type[_U]], Type[_U]]:
@@ -135,6 +141,9 @@ def dataclass(
     order: bool = False,
     unsafe_hash: bool = False,
     frozen: bool = False,
+    match_args: bool = True,
+    kw_only: bool = False,
+    slots: bool = False,
     base_schema: Optional[Type[marshmallow.Schema]] = None,
     cls_frame: Optional[types.FrameType] = None,
 ) -> Union[Type[_U], Callable[[Type[_U]], Type[_U]]]:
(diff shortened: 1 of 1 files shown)


marshmallow_dataclass/__init__.py
```diff
@@ -163,10 +172,46 @@ def dataclass(
...
+        }
+
+        for arg, arg_info in args.items():
+            if locals()[arg] is not arg_info["default_value"]:
+                raise ValueError(arg_info["error_message"])
+
+        # dataclass's typing doesn't expect it to be called as a function, so ignore type check
+        dc = dataclasses.dataclass(  # type: ignore
+            _cls, repr=repr, eq=eq, order=order, unsafe_hash=unsafe_hash, frozen=frozen
+        )
```
lovasoa, line 214: Why this ? Can't we just forward the arguments we got to the base dataclass implementation, and let it error if it wants to ?

lovasoa (changes requested):

marshmallow_dataclass/__init__.py
```diff
@@ -117,6 +120,9 @@ def dataclass(
     order: bool = False,
     unsafe_hash: bool = False,
     frozen: bool = False,
+    match_args: bool = True,
+    kw_only: bool = False,
+    slots: bool = False,
     base_schema: Optional[Type[marshmallow.Schema]] = None,
     cls_frame: Optional[types.FrameType] = None,
 ) -> Callable[[Type[_U]], Type[_U]]:
```
dairiki, line 128: I feel these overloads should be conditional on python version.  I.e. The prototypes should not declare unsupported options.

Since (I think) we want `marshmallow_dataclass.dataclass` to have the same signature as `dataclasses.dataclass`, we should probably just copy typeshed's interface definition for `dataclass`.

https://github.com/python/typeshed/blob/8b9b0f4d031fe614a0211e1faadf57b3340e643e/stdlib/dataclasses.pyi#L57C1-L104C1.

marshmallow_dataclass/__init__.py
dairiki: Agree.  For the implementation can we not just do
```py
def dataclass(
    __cls: Optional[Type[_U]] = None,
   *,
   base_schema: Optional[Type[marshmallow.Schema]] = None,
   cls_frame: Optional[types.FrameType] = None,
   **kwargs: bool,
 ) -> Union[Type[_U], Callable[[Type[_U]], Type[_U]]]:
    dc = dataclasses.dataclass(**kwargs)
    if cls_frame is None:
        # ... logic to find frame unchanged ...

    def decorate(cls: Type[_U]) -> Type[_U]:
        return add_schema(dc(cls), base_schema, cls_frame=cls_frame)

    if _cls is None:
        return decorate
    return decorate(cls)
```

dairiki (commented):
I've left a couple of comments here.

Also, see my review of #205 for more (albeit similar) comments.
https://github.com/lovasoa/marshmallow_dataclass/pull/205/files

marshmallow_dataclass/__init__.py
crutcher: This breaks the type signature.

crutcher:
I tried to address some of the comments here in another PR:
https://github.com/lovasoa/marshmallow_dataclass/pull/278

PR #260, opened 2024-05-08
Update depedency typeguard

### LINKED ISSUE #250: Union with generic types is broken
```python
from typing import Union
import marshmallow.fields as mf
from marshmallow_dataclass import NewType, dataclass
from webargs.fields import DelimitedList

DelimitedListInt = NewType('DelimitedListInt', list[int], field=lambda *args, **kwargs: DelimitedList(mf.Integer(), *args, **kwargs))

@dataclass
class Test:
  myint: Union[DelimitedListInt, list[int]]


Test.Schema().load({'myint': '1,2,3'})
```

Raises: 
```
TypeError: isinstance() argument 2 cannot be a parameterized generic
```

The issue is that `typeguard` does a check of `isinstance([1, 2, 3], list[int])` which is not allowed. 
Updating to typeguard 4.0.0 fixes the issue.  I've also verified that the latest version, 4.1.5 works.

jasle:
Resolves #250

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ EXTRAS_REQUIRE = {
         "marshmallow-enum; python_version < '3.7'",
         "marshmallow>=3.18.0,<4.0; python_version >= '3.7'",
     ],
-    "union": ["typeguard>=2.4.1,<4.0.0"],
+    "union": ["typeguard>=2.4.1,<5.0.0"],
     "lint": ["pre-commit~=2.17"],
     ':python_version == "3.6"': ["dataclasses", "types-dataclasses<0.6.4"],
     "docs": ["sphinx"],


jasle:
cc @LostInDarkMath

lovasoa:
looks like this breaks the tests

jasle:
> looks like this breaks the tests

It does not directly break the tests, but `typeguard` 4.0.0 requires atleast <3.7.4, so the building on ubuntu with python 3.6 does not work.

I adjusted the required version in this PR, but so this does not fix #250 for all specified python versions for this project.

Is dropping the support for older python versions an option?
`typeguard` 4.0.0 requires at least python 3.7.4, the current version of `marshmallow` python 3.8.

lovasoa:
Yes, that would be okay! Can you open a pr?

lovasoa (approved):

PR #261, opened 2024-05-09
Added test for union with generics

jasle:
Add test for issue #250


jasle:
This PR should currently fail the checks with older python versions.

I will remove the draft status when the support for them is dropped.r

jasle:
Added skipif, ready to merge.

lovasoa (approved):

lovasoa (approved):
we can merge this as-is with the skipIf

PR #263, opened 2024-05-09
Change supported python versions to 3.8-3.12

jasle:
Update supported python version to 3.8-3.12 and adjust checks accordingly.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -12,10 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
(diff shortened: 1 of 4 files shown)


lovasoa:
I'm ok with dropping support for python 3.6 and 3.7

Together they represent <1.5% of downloads

image

https://pypistats.org/packages/marshmallow-dataclass

But can you please bump the major version, since this is a breaking change ?

jasle:
> But can you please bump the major version, since this is a breaking change ?

Raise it to 9.0.0 right?

lovasoa (approved):

lovasoa:
Yes, this should be 9.0.0

PR #266, opened 2024-05-25
Ability to override nested class

ddorian:
I need to override the nested class. This is to use the same object in different presentations.
A simple example would be we have a `User` class.
And we want 2 presentations of it, one as `UserDetails` (this is the User.Schema by default) on an API endpoint.
The other as a `UserEvent`, which has a different schema altogether.

Is this approach ok? I can add tests etc.

### DIFF
--- a/marshmallow_dataclass/__init__.py
+++ b/marshmallow_dataclass/__init__.py
@@ -886,7 +886,8 @@ def _field_for_schema(
     forward_reference = getattr(typ, "__forward_arg__", None)
 
     nested = (
-        nested_schema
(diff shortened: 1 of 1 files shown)


dairiki:
Maybe I'm misunderstanding your goal here, but `marshmallow_field` can already be used to specify an explicit schema field for a particular dataclass field. 

```py
@dataclass
class Parent:
    user: User = field(metadata={marshmallow_field=marshmallow.field.Nested(UserEventSchema)})
```

ddorian:
You're correct! Thank you!

PR #284, opened 2025-09-09
chore(python): drop support for Python3.8

noirbee:
mypy dropped support for it in v1.17, and it has been EOL since October 2024.

cf
* https://github.com/python/mypy/blob/master/CHANGELOG.md#remove-support-for-targeting-python-38
* https://devguide.python.org/versions/


noirbee:
@lovasoa This is more or less needed in order to merge https://github.com/lovasoa/marshmallow_dataclass/pull/286; would you please review / merge it ?

lovasoa (approved):

PR #287, opened 2026-03-02
Test on Python 3.13 and 3.14

cjwatson:


cjwatson:
Getting this to work as the current matrix approach stands is impossible.  pyupgrade 3.21.1 includes a fix needed for Python 3.14.1, but the same release also drops support for Python 3.9.

I'd suggest that `pre-commit` should be a separate CI job, run on fewer versions of Python (perhaps only one, but you'd have to tell me the correct policy to apply there).  It doesn't need to be run using every version of Python that the package itself supports.

lovasoa:
agreed ! pre-commit should run only on a single widely used modern python (let's say 3.14), while the library tests themselves should run on the full matrix

cjwatson:
OK, that seems to work now.  You could probably also do a more general `pre-commit autoupdate` or similar (or even add something like pre-commit.ci or https://pre-commit.ci/lite.html to automate that), but I decided to keep my changes reasonably minimal for now.

lovasoa (approved):
thank you !

